PR #238, opened 2020-07-04
Add support for phasing linked reads

marcelm:
(Transferred from https://bitbucket.org/whatshap/whatshap/pull-requests/43)

This adds preliminary support for treating read clouds (reads that all have the same BX tag value) as a single read. This allows us to phase data from 10X Chromium and similar technologies.

Use `--linked-reads` to enable.

I just finished this and ran it on my small test dataset, but it needs more testing. Although we’ve previously discussed that physical coverage could be a problem, I don’t see that on this small test dataset. In any case, the results are definitely better than when not using `--linked-reads`.

@tobiasmarschall commented:
> When we tried this years ago (see branch 10xG_phasing), the results weren’t horrible, but worse than LongRanger, particularly in terms of switch errors.

### DIFF
--- a/tests/test_readsetreader.py
+++ b/tests/test_readsetreader.py
@@ -1,7 +1,7 @@
 import pytest
 
 from whatshap.core import Read, Variant
-from whatshap.variants import merge_two_reads, merge_reads
+from whatshap.variants import merge_two_reads, merge_reads, ReadSetReader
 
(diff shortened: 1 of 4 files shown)


marcelm:
I’ve rebased this PR on top of main in case anyone wants to try this.

PR #239, opened 2020-07-24
Haploid phasing blocks

schrins:
I added a new parameter for whatshap polyphase, with which the phasing set information can be included into the VCF on a haplotype level. In order to do this, a new custom format field "HS" was added the VCF writer and a list of values (one for each phase) is added. This feature was promised in a recent submission about the polyploid phaser of whatshap, so it needs to be added at some point.

Disclaimer 1: I am always a bit insecure when editing the vcf.py, so someone should review whether I violated any conventions.
Disclaimer 2: A lot of files were reformatted after running black. I needed to do this, because otherwise the test builds with black would fail. I remember that the unit tests were not supposed to be reformatted, but since tox internally runs "black --check whatshap/ tests/ setup.py" I assumed, black has to be run on them as well.

### DIFF
--- a/whatshap/cli/polyphase.py
+++ b/whatshap/cli/polyphase.py
@@ -92,6 +92,7 @@ def run_polyphase(
     indels=True,
     mapping_quality=20,
     tag="PS",
+    include_haploid_sets=False,
     write_command_line_header=True,
     read_list_filename=None,
     ce_bundle_edges=False,
@@ -148,6 +149,7 @@ def run_polyphase(
                     out_file=output,
                     tag=tag,
                     ploidy=ploidy,
+                    include_haploid_sets=include_haploid_sets,
                 )
             )
         except OSError as e:
@@ -155,7 +157,7 @@ def run_polyphase(
 
         vcf_reader = stack.enter_context(
             VcfReader(
-                variant_file, indels=indels, phases=True, genotype_likelihoods=False, ploidy=ploidy,
+                variant_file, indels=indels, phases=True, genotype_likelihoods=False, ploidy=ploidy
             )
         )
 
@@ -222,7 +224,7 @@ def run_polyphase(
                     continue
 
                 # These two variables hold the phasing results for all samples
-                superreads, components = dict(), dict()
+                superreads, components, haploid_components = dict(), dict(), dict()
 
                 # Iterate over all samples to process
                 for sample in samples:
@@ -250,14 +252,13 @@ def run_polyphase(
                         len(missing_genotypes),
                     )
                     logger.info(
-                        "Number of remaining heterozygous variants: %d",
-                        len(phasable_variant_table),
+                        "Number of remaining heterozygous variants: %d", len(phasable_variant_table)
                     )
 
                     # Get the reads belonging to this sample
                     timers.start("read_bam")
                     readset, vcf_source_ids = phased_input_reader.read(
-                        chromosome, phasable_variant_table.variants, sample,
+                        chromosome, phasable_variant_table.variants, sample
                     )
                     readset.sort()
                     timers.stop("read_bam")
@@ -300,7 +301,7 @@ def run_polyphase(
 
                         # Re-read the readset to remove discarded variants
                         readset, vcf_source_ids = phased_input_reader.read(
-                            chromosome, phasable_variant_table.variants, sample,
+                            chromosome, phasable_variant_table.variants, sample
                         )
                         readset.sort()
                         timers.stop("verify_genotypes")
@@ -309,25 +310,31 @@ def run_polyphase(
                     readset = readset.subset(
                         [i for i, read in enumerate(readset) if len(read) >= max(2, min_overlap)]
                     )
-                    logger.info(
-                        "Kept %d reads that cover at least two variants each", len(readset),
-                    )
+                    logger.info("Kept %d reads that cover at least two variants each", len(readset))
 
                     # Adapt the variant table to the subset of reads
                     phasable_variant_table.subset_rows_by_position(readset.get_positions())
 
                     # Run the actual phasing
-                    sample_components, sample_superreads = phase_single_individual(
-                        readset, phasable_variant_table, sample, phasing_param, output, timers,
+                    (
+                        sample_components,
+                        sample_haploid_components,
+                        sample_superreads,
+                    ) = phase_single_individual(
+                        readset, phasable_variant_table, sample, phasing_param, output, timers
                     )
 
                     # Collect results
                     components[sample] = sample_components
+                    haploid_components[sample] = sample_haploid_components
                     superreads[sample] = sample_superreads
 
                 with timers("write_vcf"):
                     logger.info("======== Writing VCF")
-                    vcf_writer.write(chromosome, superreads, components)
+                    if include_haploid_sets:
+                        vcf_writer.write(chromosome, superreads, components, haploid_components)
+                    else:
+                        vcf_writer.write(chromosome, superreads, components)
                     # TODO: Use genotype information to polish results
                     # assert len(changed_genotypes) == 0
                     logger.info("Done writing VCF")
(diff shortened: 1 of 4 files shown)


marcelm:
The Black issue is a bit strange. For me, `tox -e black` succeeds. Which problem did you encounter? One problem could be that you used an old Black version (it should be 19.10b0).

Also, at the moment, it’s not possible to see which changes you made and which were made by Black. In general, if you run an automated tool on the source code, those changes should be in a separate commit. In this case, I’d even prefer it to be a separate PR, but perhaps we can figure this out without needing to reformat at all.

schrins:
I actually had a newer version locally installed (19.30b0), which formatted the code in such a way, that version 19.10b0 complained about it.

I ran "black whatshap/ tests/ setup.py", which probably changed too many files. I think the easiest thing I could do is to go back to the master, then check for figuring out how to make tox run and then re-appling the actual changes manually (I changed only three files or so). I will purge the current history of the branch and only include the new commits.

schrins:
I removed all unnecessary reformats and  squashed everything into one commit.

The issues with black were indeed caused by my installed black version being different from the one tox uses. Before any reformats, tox complained about some files (the actually edited ones) not being formatted correctly. I ran the "black --check whatshap/ tests/ setup.py" command without the "--check" with my local black version, which then changed a ton of untouched files as well (interestingly, after this there was still one wrong formatting remaining, because 19.10b0 did not like, what 19.30b0 produced :) ). Now everything should be fine.

schrins:
What are we going to do with this PR? Do we want to have this feature in the main? Any objections to the changes in vcf.py?

marcelm:
I’m still not back at work, but let me take a look.

schrins:
If you are on vacation, it's fine. I just wanted to remind anyone, that it's still open :)

whatshap/cli/polyphase.py
```diff
@@ -981,6 +1002,12 @@ def add_arguments(parser):
         help="Ignore read groups in BAM/CRAM header and assume all reads come "
         "from the same sample.",
     )
+    arg(
+        "--include-haploid-sets",
+        default=False,
+        action="store_true",
+        help="Includes the phase set information for every single haplotype in a custom VCF format field 'HS'.",
```
marcelm, line 1009:
```suggestion
        help="Include the phase set information for every single haplotype in a custom VCF format field 'HS'.",
```

marcelm: I’d also like to see some documentation for new features, but we haven’t really enforced this in the past, so it would be ok for me to omit it now. But generally, I think we should avoid that the only documentation for a feature is the help text on the command line (except in obvious cases).

whatshap/cli/polyphase.py
```diff
@@ -148,14 +149,15 @@ def run_polyphase(
...
                     ploidy=ploidy,
+                    include_haploid_sets=include_haploid_sets,
                 )
             )
         except OSError as e:
             raise CommandLineError(e)
 
         vcf_reader = stack.enter_context(
             VcfReader(
-                variant_file, indels=indels, phases=True, genotype_likelihoods=False, ploidy=ploidy,
```
marcelm, line 158: The trailing command should not be removed. (There are still some extraneous formatting changes in this PR; I won’t remark on the others.)

whatshap/cli/polyphase.py
```diff
@@ -309,25 +310,31 @@ def run_polyphase(
...
+                    haploid_components[sample] = sample_haploid_components
                     superreads[sample] = sample_superreads
 
                 with timers("write_vcf"):
                     logger.info("======== Writing VCF")
-                    vcf_writer.write(chromosome, superreads, components)
+                    if include_haploid_sets:
+                        vcf_writer.write(chromosome, superreads, components, haploid_components)
+                    else:
+                        vcf_writer.write(chromosome, superreads, components)
```
marcelm, line 337: A single `.write` call would avoid some code duplication:
```
vcf_writer.write(chromosome, superreads, components, haploid_components if include_haploid_sets else None)
```

whatshap/cli/polyphase.py
```diff
@@ -449,35 +447,51 @@ def phase_single_individual(readset, phasable_variant_table, sample, phasing_par
...
     accessible_positions = sorted(readset.get_positions())
     components = {}
+    haploid_components = {}
 
     ext_cuts = cut_positions + [num_vars]
     for i, cut_pos in enumerate(cut_positions):
         for pos in range(ext_cuts[i], ext_cuts[i + 1]):
             components[accessible_positions[pos]] = accessible_positions[ext_cuts[i]]
             components[accessible_positions[pos] + 1] = accessible_positions[ext_cuts[i]]
+            haploid_components[accessible_positions[pos]] = [0 for _ in range(phasing_param.ploidy)]
```
marcelm, line 482: Use `[0] * phasing_param.ploidy` to create this list (and in the next line as well)

whatshap/threading.py
```diff
@@ -48,7 +45,12 @@ def run_threading(readset, clustering, ploidy, genotypes, block_cut_sensitivity)
     # we can look at the sequences again to use the most likely continuation, when a haplotype leaves a collapsed cluster (currently inactive)
     path = improve_path_on_collapsedswitches(path, num_clusters, c_to_c_global)
 
+    # determine cut positions
```
marcelm, line 48: That comment is redundant (the function name `compute_cut_positions` already says the same thing).

whatshap/threading.py
```diff
@@ -191,7 +199,13 @@ def compute_cut_positions(path, block_cut_sensitivity, num_clusters):
             if dissim >= dissim_threshold:
                 cpn_rising = [False for c_id in range(num_clusters)]
```
marcelm, line 200: Not part of this PR, but this could also become `[False] * num_clusters`.

whatshap/threading.py
```diff
@@ -191,7 +199,13 @@ def compute_cut_positions(path, block_cut_sensitivity, num_clusters):
...
                 cut_positions.append(i)
-    return cut_positions
+
+                # get all cut threads
+                threads_cut = [j for j in range(ploidy) if path[i - 1][j] in clusters_cut]
+                for thread in threads_cut:
+                    haploid_cut_positions[thread].append(i)
```
marcelm, line 206: Just as a suggestion for later: The code here is getting indented quite deeply. It would be nice to refactor this at some point.

whatshap/vcf.py
```diff
@@ -963,7 +984,18 @@ def write(self, chromosome: str, sample_superreads, sample_components):
                         )
                         is_het = not genotypes[pos].is_homozygous()
 
-                    if pos in components and pos in phases and is_het:
+                    if (
```
marcelm, line 987: A couple of comments earlier, I made a suggestion of how to avoid the duplicate call to `VcfWriter.write`. I suggest to do something similar here: Call `self._set_phasing_tags` only once with four parameters, but use an `if` before that call to set the fourth parameter.

Also, list the `pos in components and pos in phases and is_het` conditionals only once. The `if haploid_components ...` can be in a nested if. (Not sure if that explanation is sufficient to understand what I mean, just let me know.)

schrins:
I pushed the requested changes, except the one for documentation. Actually, I still have an uncommited guide.rst an my system, where basic polyphase functionality is explained for the main page. I think this should be discussed in another issue.

whatshap/vcf.py
```diff
@@ -964,7 +987,16 @@ def write(self, chromosome: str, sample_superreads, sample_components):
...
-                        self._set_phasing_tags(call, components[pos], phases[pos])
+                        self._set_phasing_tags(
+                            call,
+                            components[pos],
+                            phases[pos],
+                            haploid_components[pos]
+                            if haploid_components
+                            and pos in haploid_components
+                            and len(haploid_components[pos]) == self.ploidy
+                            else None,
```
marcelm, line 998: Embedding the conditional expression on multiple lines looks a bit weird and is a bit hard to read. Can you pull this out as a regular `if` that sets a temporary variable (`haploid_component` or so)?

schrins: I changed the statement and also rebased with the current master.

whatshap/vcf.py
```diff
@@ -964,7 +987,17 @@ def write(self, chromosome: str, sample_superreads, sample_components):
...
+                            haploid_components
+                            and pos in haploid_components
+                            and len(haploid_components[pos]) == self.ploidy
+                        )
+                        if use_haploid_components:
+                            self._set_phasing_tags(
+                                call, components[pos], phases[pos], haploid_components[pos]
+                            )
+                        else:
+                            self._set_phasing_tags(call, components[pos], phases[pos], None)
```
marcelm, line 1000: This is what I meant (avoids writing out two calls to `self._set_phasing_tags`):
```suggestion
                        haploid_component = haploid_components[pos] if (
                            haploid_components
                            and pos in haploid_components
                            and len(haploid_components[pos]) == self.ploidy
                        ) else None
                        self._set_phasing_tags(
                            call, components[pos], phases[pos], haploid_component
                        )
```

schrins: Ah, sorry I was not sure what you meant with the temporary variable. I pushed another version (and squashed it with the previous commit).

marcelm: Thanks, looks good now. Entirely my fault of not conveying what I wanted from the beginning :-).

(The reformatting that black has done now makes me wonder whether it would look better with an explicit `if (...): haploid_component = ...; else: haploid_component = None`, that is, not with an if expression, but let’s just keep this version now.)

marcelm (approved):

marcelm:
Looks good now! Please merge if you think it’s ready.

Issue #251, opened 2020-09-23
Haplotag fails when writing to stdout


pontushojer:
I am using `whatshap haplotag` to assign haplotypes to linked-read data. 

The following command works without any issues:

```
$ whatshap haplotag my.phased.vcf.gz my.bam --linked-read-distance-cutoff 30000 --reference ref.fasta --ignore-read-groups  -o my.haplotag.bam 
Found 1 samples in input VCF
Keeping 1 samples for haplo-tagging
Found 1 samples in BAM file
Reading alignments and detecting alleles ...
Found 41 reads covering 5 variants
Reading alignments and detecting alleles ...
Found 0 reads covering 0 variants
Reading alignments and detecting alleles ...
Found 0 reads covering 0 variants
Reading alignments and detecting alleles ...
Found 0 reads covering 0 variants

== SUMMARY ==
Total alignments processed:                      8380
Alignments that could be tagged:                  725
Alignments spanning multiple phase sets:            0
haplotag - total processing time: 0.11714005470275879
```

But when output is directed to stdout the program just stalls. See command below: 

```
$ whatshap haplotag my.phased.vcf.gz my.bam --linked-read-distance-cutoff 30000 --reference ref.fasta --ignore-read-groups > my.haplotag.bam
Found 1 samples in input VCF
Keeping 1 samples for haplo-tagging
Found 1 samples in BAM file
```

To stop it I had to kill the process.

aalsabag:
Hi @pontushojer ,
I just took a look at the code and noticed that the default value for the output file is `None`.
I made a slight code change to make it actually default to stdout and raised a [PR](https://github.com/whatshap/whatshap/pull/253). I think it is working as it should
The following command continues to work without issue:
```
$ whatshap haplotag tests/data/pacbio/phased.vcf.gz tests/data/pacbio/pacbio.bam --linked-read-distance-cutoff 30000 --reference tests/data/pacbio/reference.fasta --ignore-read-groups -o ./test1.bam
Found 1 samples in input VCF
Keeping 1 samples for haplo-tagging
Found 1 samples in BAM file
Reading alignments and detecting alleles ...
Found 25 reads covering 54 variants

== SUMMARY ==
Total alignments processed:                        25
Alignments that could be tagged:                   25
Alignments spanning multiple phase sets:            0
haplotag - total processing time: 0.23807978630065918
```

And now also this command works without hanging:
```
[...]

PR #253, opened 2020-09-27
Issue 251

aalsabag:
As per issue #251 . Let me know if this is wrong. I'd be happy to change whatever is necessary.


tobiasmarschall:
Thanks for this PR. Comparing to the `phase` subcommand, I notice that we've used `sys.stdout` there: https://github.com/whatshap/whatshap/blob/28001d90a2238479a9d7b99502f88c6d897d6fb9/whatshap/cli/phase.py#L953 
which I think is preferable. @marcelm do you agree?

aalsabag:
Silly me, I forgot to consider OS's other than linux. 🤦 
Thank you
updated

Issue #207, opened 2019-09-20
Allow `--chromosome` to accept a comma-delimited list of chromosomes.


marcelm:
**[Original report](https://bitbucket.org/whatshap/whatshap/issue/207) by William Rowell (Bitbucket: [williamrowell](https://bitbucket.org/williamrowell), GitHub: [williamrowell](https://github.com/williamrowell)).**

----------------------------------------

I would like to calculate phase block size only for autosomes in a male, and while I can use tabix to remove the sex chromosomes, it would be convenient to pass a list of autosomes to whatshap stats.  Would it be possible to allow the `--chromosome` argument for `whatshap phase` and/or `whatshap stats` to accept a comma delimited list as an argument?

asan-emirsaleh:
There is still issue with `whatshap phase`: when one specify several chromosome names via `--chromosome` key, the pipeline returns error
```
whatshap phase: error: argument --chromosome: expected one argument
Failed to read from standard input: unknown file type
```

So there is no a useful way of specifying multiple chromosome names to process in `whatshap`, as well to parallelize computation.

Best regards
Asan

marcelm:
Which exact command line did you use?

PR #258, opened 2020-10-09
Allow multiple `--chromosome` arguments in `whatshap stats` (#207)

pontushojer:
This implementation is similar to how it was done in e.g. `whatshap phase`. While this does not fix the requested comma-separate input of chromosomes in #207, it achieves the same thing.


marcelm:
That looks good to me, thanks!

PR #261, opened 2020-10-16
Added description for the polyphase module.

schrins:
Should fix issue 260. Provides some detailed information regarding the polyphase module.

### DIFF
--- a/doc/guide.rst
+++ b/doc/guide.rst
@@ -770,11 +770,8 @@ Therefore the output genotypes can be different than the input genotypes.
 
 3. Polyploid phasing on pedigrees is not supported yet.
 
-4. Structural variants are not considered during the phasing. Long indels can 
-cause some variants to not be present on all haplotypes. As of now, these 
-variants are not handled by the phasing algorithm. Thus, it will always output 
-as many haplotypes on all variants as were specified by the command line 
-argument.
+4. The phasing algorithm does not consider copy number variants and always 
+produces the provided number of haplotypes at any location.
 
 There is no strict limitation regarding the coverage of the input reads. 
 However, the running time grows quadratically with the coverage. For that
--- a/doc/guide.rst
+++ b/doc/guide.rst
@@ -750,3 +750,53 @@ an input to the above mentioned genotyping commands. This can be done by running
     whatshap find_snv_candidates ref.fasta input.bam -o variants.vcf
 
 If Nanopore reads are used for calling SNPs, it is recommended to add option --nanopore to the above command.
+
+Polyploid Phasing
+=================
+
+In addition to diploid phasing, WhatsHap also supports polyploid phasing 
+through a different algorithm. The ``whatshap polyphase`` command works 
+almost the same as the ``phase`` command with a few restrictions:
+
+1. An additional integer argument ``--ploidy`` must be specified. This ploidy
+must match the ploidy in the provided VCF file(s). The ploidy also greatly
+impacts the running time as the phasing becomes more complex. Ploidies
+higher than 6 may take very long to process.
+
+2. WhatsHap will use available genotype information from the VCF file(s), but
+the computed haplotypes are not guaranteed to follow these genotypes, if they
+deviate too much from the allele distribution among the aligned reads.
+
+3. Polyploid phasing on pedigrees is not supported yet.
+
+4. Structural variants are not considered during the phasing. The algorithm
+will always produce the provided number of haplotypes at any location.
+
+Opposed to the diploid phasing algorithm, there is no strict limitation
+regarding the coverage of the input reads. However, it still increases the
+running time and we do not recommend to use more than 120X. In principle it is
+possible to phase diploid samples via the ``polyphase`` command, but the
+results will likely be less accurate than the diploid phasing mode, as the
+latter is more specialized for the diploid case.
+
+A problem, which occurs in polyploid phasings is the connectivity of phased
+variants. The more haplotypes exist the more reads are required between
+consecutive variants to properly connect haplotype fragments from both sides.
+As a result, every phased variant will receive a phased block ID, such that
+all variants with the same ID belong to the same haplotype block. By default
+WhatsHap is very conservative with these blocks and splits them, if it could 
+not resolve ambiguity between consecutive variants. This behavior can be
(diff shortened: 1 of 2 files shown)


doc/guide.rst
```diff
@@ -750,3 +750,53 @@ an input to the above mentioned genotyping commands. This can be done by running
...
+=================
+
+In addition to diploid phasing, WhatsHap also supports polyploid phasing 
+through a different algorithm. The ``whatshap polyphase`` command works 
+almost the same as the ``phase`` command with a few restrictions:
+
+1. An additional integer argument ``--ploidy`` must be specified. This ploidy
+must match the ploidy in the provided VCF file(s). The ploidy also greatly
+impacts the running time as the phasing becomes more complex. Ploidies
+higher than 6 may take very long to process.
```
tobiasmarschall, line 764: Can we be more specific here and possibly mention example runtimes?

doc/guide.rst
```diff
@@ -750,3 +750,53 @@ an input to the above mentioned genotyping commands. This can be done by running
...
+
+2. WhatsHap will use available genotype information from the VCF file(s), but
+the computed haplotypes are not guaranteed to follow these genotypes, if they
+deviate too much from the allele distribution among the aligned reads.
+
```
tobiasmarschall, line 769: So WhatsHap changes the genotypes, right? So maybe we should explicitly say that the genotypes in the output file can be different from the input file.

doc/guide.rst
```diff
@@ -750,3 +750,53 @@ an input to the above mentioned genotyping commands. This can be done by running
...
+3. Polyploid phasing on pedigrees is not supported yet.
+
+4. Structural variants are not considered during the phasing. The algorithm
+will always produce the provided number of haplotypes at any location.
+
```
tobiasmarschall, line 774: What does structural variant mean hear? Does it do indels?

doc/guide.rst
```diff
@@ -750,3 +750,53 @@ an input to the above mentioned genotyping commands. This can be done by running
...
+Opposed to the diploid phasing algorithm, there is no strict limitation
+regarding the coverage of the input reads. However, it still increases the
```
tobiasmarschall, line 776: I think this statement on the diploid algorithm is confusing and would remove it.

doc/guide.rst
```diff
@@ -750,3 +750,53 @@ an input to the above mentioned genotyping commands. This can be done by running
...
+running time and we do not recommend to use more than 120X. In principle it is
+possible to phase diploid samples via the ``polyphase`` command, but the
+results will likely be less accurate than the diploid phasing mode, as the
+latter is more specialized for the diploid case.
+
+A problem, which occurs in polyploid phasings is the connectivity of phased
+variants. The more haplotypes exist the more reads are required between
+consecutive variants to properly connect haplotype fragments from both sides.
```
tobiasmarschall, line 784:
```suggestion
To achieve reliable phasing, as many haplotypes as possible should be represented
in the input reads. In case of unrepresented haplotypes, phasing can become impossible
and the output haplotypes are broken into phased blocks.
```

doc/guide.rst
```diff
@@ -750,3 +750,53 @@ an input to the above mentioned genotyping commands. This can be done by running
...
+As a result, every phased variant will receive a phased block ID, such that
+all variants with the same ID belong to the same haplotype block. By default
+WhatsHap is very conservative with these blocks and splits them, if it could
```
tobiasmarschall, line 787:
```suggestion
WhatsHap is very conservative with these blocks and splits them whenever it could 
```

doc/guide.rst
```diff
@@ -750,3 +750,53 @@ an input to the above mentioned genotyping commands. This can be done by running
...
+variants. The more haplotypes exist the more reads are required between
+consecutive variants to properly connect haplotype fragments from both sides.
+As a result, every phased variant will receive a phased block ID, such that
+all variants with the same ID belong to the same haplotype block. By default
+WhatsHap is very conservative with these blocks and splits them, if it could 
+not resolve ambiguity between consecutive variants. This behavior can be
+adjusted via the ``--block-cut-sensitivity`` parameter. Valid values range from
+0 to 5 (including) with a default of 4. A lower sensitivity will produce longer
+phasing blocks, which might contain more switch errors, though. A sensitivity
+of 1 means that haplotypes are only cut, where there was no read connecting two
```
tobiasmarschall, line 792:
```suggestion
of 1 means that haplotypes are only cut at positions where there was no read connecting two
```

doc/guide.rst
```diff
@@ -750,3 +750,53 @@ an input to the above mentioned genotyping commands. This can be done by running
...
+consecutive variants.
```
tobiasmarschall, line 793:
```suggestion
consecutive variants (in any haplotype).
```

tobiasmarschall (commented):
Thanks a lot for adding this. See some few suggestions below. Could you also add the citation to the corresponding doc page?

doc/guide.rst
schrins: We only have example runtimes for our multi-human sample. Could be a bit confusing to state them here, because this is not what you would usually phase with polyphase.

doc/guide.rst
schrins: I think we never tested indels, but only SNPs. Structural variant means long indels in which not all haplotypes are present. The threading will always output --ploidy many haplotypes at all times. For us this is clear, but users might not expect it. I can  write this part in more detail.

tobiasmarschall: Ok, how about this to clarify:

```suggestion

4. The phasing algorithm does not consider copy number variants and always produces the provided number of haplotypes at any location.

```

tobiasmarschall: I'd use the term copy number variants instead of "long indels" or "structural variants" here...

schrins: I'm fine with it. Changed now.

schrins:
I think the comments are almost resolved. The running time example is still unclear. I added the Genome Biology citation, but did not delete the arXiv citation. But it would make sense, because we don't want people to cite the preprint, do we?

tobiasmarschall:
Yes, the biorxiv citation can be remove now.

tobiasmarschall:
Thanks, @schrins !

Issue #262, opened 2020-10-19
extracting reads from (multiple) regions


dcopetti:
Hello,

I would like to tag (and then extract separately) long reads based on the haplotypes that I built with longranger wgs for about 900 regions in my assembly. I can't figure out a way to specify the coordinates: if I do it as `OF1p3c_SSc_Bd_SSc_all_1:320000-520000` like the documentation says, I get `ValueError: invalid literal for int() with base 10: '320000-520000'`. If I remove the coordinates and leave only sequence name
`whatshap haplotag --ignore-read-groups --reference reference.fa --regions OF1p3c_SSc_Bd_SSc_all_1 phased_variants.vcf.gz long_reads_aligned.bam`
it works. 

I tried also having `--regions` file with the regions separated by space:
`OF1p3c_SSc_Bd_SSc_all_1:320000-520000 OF1p3c_SSc_Bd_SSc_all_1:770000-970000 OF1p3c_SSc_Bd_SSc_all_1:5570000-5770000`
but it does not accept a file like that How should I encode those intervals?
As a suggestion, I think it would be great if the user could specify the regions in a bed file, for compatibility with other steps.
Thanks,
Dario

tobiasmarschall:
As far as I remember, @ptrebert refactored the region parsing, so I hope he might be able to help.

ptrebert:
It looks like the relevant part of the `haplotag` module has been rewritten again; however, the code looks good to me, and I cannot reproduce the above `ValueError` with the current code base:
```
Region.parse("OF1p3c_SSc_Bd_SSc_all_1:320000-520000 ")
Region("OF1p3c_SSc_Bd_SSc_all_1", 319999, 520000)
```
What version of WhatsHap are you using?

dcopetti:
Hi Peter,
I am using whatshap 1.0.
The command above completed overnight, it printed out this
Found 1 samples in input VCF
Keeping 1 samples for haplo-tagging
Found 0 samples in BAM file
WARNING: Skipping duplicated position 71252637 on chromosome 'OF1p3c_SSc_Bd_SSc_all_1'
Reading alignments and detecting alleles ...
Found 48034 reads covering 62114 variants

== SUMMARY ==
Total alignments processed:                    131049
Alignments that could be tagged:                44757
Alignments spanning multiple phase sets:         1321
haplotag - total processing time: 68.04450273513794
and a bam file where I can not find the phase assignment for each read:
[...]

PR #264, opened 2020-10-21
Allow colon in region specifications

marcelm:
For backwards compatibility, we need to allow `chr:start:end` in addition to `chr:start-end` when specifying a `--region` for `whatshap haplotag`.

I have also simplified the tests a bit. It’s not necessary to run the full `run_haplotag` function just to see whether an invalid region is correctly caught, we should instead test the parsing code directly.

See #262

### DIFF
--- a/tests/test_run_haplotag.py
+++ b/tests/test_run_haplotag.py
@@ -3,7 +3,7 @@ import shutil
 import pysam
 import pytest
 
-from whatshap.cli.haplotag import run_haplotag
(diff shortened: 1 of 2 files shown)


marcelm:
By the way, I have previously pushed some minor changes to `master` directly, but that isn’t good practice since it bypasses CI. I’ll therefore spend PRs even for smaller changes in the future. If I consider them to be trivial, I can still merge them myself if I don’t get a review.

PR #265, opened 2020-10-21
Parallel polyphase

schrins:
This pull request contains a parallelization for the polyphase module. The algorithms themselves are not changed, but the phasing is parallelized on a block-level. This means that it can only achieve a speed-up, if there are actually some weekly (or un-)connected components in the readset. The implementation uses the `multiprocessing` library of Python, which forks the main process. A thread-level parallelization is technically not possible with Python due the Global Interpreter Lock.

The parallelization must be requested via the `--threads` parameter when executing the module. The default is 1 thread, which invokes no fork at all. Running multiple threads increases the memory consumption (see example below), because the worker threads create local data structures, which are needed for the phasing. The main readset is not duplicated per thread.

Note, that only the phasing stage is parallelized. The pre- and postprocessing (splitting blocks, writing VCF) and the I/O at the beginning are not. Therefore the overall speed-up is not too impressing, but still better than nothing. I ran one of the tetraploid-human samples from the polyphase publication (full chr1, 40X, default `whatshap polyphase` settings) with 1 thread and with 8 threads on an 8-core CPU. The running times look like this:

```
== SUMMARY ==
Maximum memory usage: 1.178 GB
Time spent reading BAM/CRAM:                  473.8 s
Time spent parsing VCF:                         4.6 s
Time spent detecting blocks:                   15.0 s
Time spent scoring reads:                      12.3 s
Time spent solving cluster editing:           286.2 s
Time spent threading haplotypes:              101.5 s
Time spent writing VCF:                         5.6 s
Time spent on rest:                             9.6 s
Total elapsed time:                           908.5 s

== SUMMARY ==
Maximum memory usage: 1.130 GB (actually about 2.75GB, manual readout with htop)
Time spent reading BAM/CRAM:                  468.5 s
Time spent parsing VCF:                         4.4 s
Time spent detecting blocks:                   15.3 s
Time spent phasing blocks:                     63.2 s
Time spent writing VCF:                         5.5 s
Time spent on rest:                             9.6 s
Total elapsed time:                           566.6 s

```

The phasing time is reduced from about 400s to 63s, which is a speed-up of ~6,4. In multithreading mode, the individual steps (scoring, cluster editing, threading) are not timed anymore, because this would require to collect timer data over all processes. I did not look into this for now, therefore there is just a total phasing time at the moment. The memory consumption is also hard to read. The current method only measures the main process, but not all child processes. I am not sure, how to solve this problem at the moment. Obviously, the memory consumption is far less than [main process consumption] times [number of threads], as the manual readout from htop shows.

I compared several phasings, both done with single und multi threading and I always got equal results, i.e. `whatshap compare` did not find a single switch or flip error between the output VCFs.

One technical note: Forking processes requires all shared objects to be (de-)serializable. I implemented a `__getstate__` and `__setstate__` method for Read and ReadSet, as suggested by the pickle documentation: https://docs.python.org/3.6/library/pickle.html#pickling-class-instances

Any thoughts about this?

### DIFF
--- a/whatshap/cli/__init__.py
+++ b/whatshap/cli/__init__.py
@@ -188,9 +188,9 @@ class PhasedInputReader:
         return readset, vcf_source_ids
 
 
-def log_memory_usage(multiprocessing=False):
+def log_memory_usage(include_children=False):
     if sys.platform == "linux":
-        if multiprocessing:
+        if include_children:
             memory_kb = (
                 resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
                 + resource.getrusage(resource.RUSAGE_CHILDREN).ru_maxrss
--- a/whatshap/cli/__init__.py
+++ b/whatshap/cli/__init__.py
@@ -188,7 +188,13 @@ class PhasedInputReader:
         return readset, vcf_source_ids
 
 
-def log_memory_usage():
+def log_memory_usage(multiprocessing=False):
     if sys.platform == "linux":
-        memory_kb = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
+        if multiprocessing:
+            memory_kb = (
+                resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
+                + resource.getrusage(resource.RUSAGE_CHILDREN).ru_maxrss
+            )
+        else:
+            memory_kb = resource.getrusage(resource.RUSAGE_SELF).ru_maxrss
         logger.info("Maximum memory usage: %.3f GB", memory_kb / 1e6)
--- a/whatshap/cli/polyphase.py
+++ b/whatshap/cli/polyphase.py
@@ -511,7 +511,7 @@ def phase_single_individual(readset, phasable_variant_table, sample, phasing_par
                 pool.apply_async(
                     phase_single_block_mt,
                     (
-                        pythonize_readset(block_readsets[block_id]),
+                        block_readsets[block_id],
                         genotype_slices[block_id],
                         phasing_param,
                         timers,
@@ -667,34 +667,6 @@ def split_readset(readset, ext_block_starts, index):
     return block_readsets
 
 
-def pythonize_readset(readset):
-    """
-    Converts a readset into a native Python data structure, which can be copied between multiple
-    CPU processes. Is only needed, when doing multithreading on the phase_single_block
-    function.
-    """
-    readlist = []
-    for read in readset:
-        readlist.append(
-            (read.name, read.source_id, [(var.position, var.allele, var.quality) for var in read])
-        )
-    return readlist
-
-
-def unpythonize_readset(python_readset):
-    """
-    Unpacks a pythonized readset back into a cython ReadSet. Can only be used in combination
-    with pythonized_readset().
-    """
-    readset = ReadSet()
-    for read in python_readset:
-        cread = Read(name=read[0], source_id=read[1])
-        for var in read[-1]:
-            cread.add_variant(var[0], var[1], var[2])
-        readset.add(cread)
-    return readset
-
-
 def phase_single_block(block_readset, genotype_slice, phasing_param, timers):
     """
     Takes as input data the reads from a single (pre-computed) block and the genotypes for all variants inside the block.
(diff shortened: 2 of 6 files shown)


marcelm:
Cool! Perhaps I have time for a deeper look, but here are a couple of thoughts.

> Note, that only the phasing stage is parallelized. Some preprocessing (splitting blocks, writing VCF) and the I/O at the beginning are not.

If parallelization were done per chromosome, it would actually be possible to do almost everything including preprocessing and even writing of the VCF file in the subprocess. The VCF and BAM need to be indexed for that, but this requirement would be worth it. Some complications arise because one would need to merge the resulting VCFs in the correct order into the final output file, but also one could skip pickling and unpickling all the data (it would be enough to just send a chromosome name to the subprocess). I’m not saying you need to rewrite this or that it’s necessarily better, but that’s probably what I would recommend if someone would want to parallelize `whatshap phase`.

> The memory consumption is also hard to read. The current method only measures the main process, but not all child processes. I am not sure, how to solve this problem at the moment.

This should be solvable by using `resource.RUSAGE_BOTH` instead of `resource.RUSAGE_SELF` in the `log_memory_usage` function at https://github.com/whatshap/whatshap/blob/491ec8e4b3bfc5624869dd7c09524769fc3b8dcd/whatshap/cli/__init__.py#L193 . The constant isn’t available on all platforms, though, so perhaps that doesn’t always work.

> A thread-level parallelization is technically not possible with Python due the Global Interpreter Lock.

To be precise: It’s possible, but probably not much faster than the unparallelized version. I would expect some speedups, though, because one can release the GIL when calling into the C++ code. When the GIL is released, threads can run in parallel even in Python.

schrins:
> If parallelization were done per chromosome, it would actually be possible to do almost everything including preprocessing and even writing of the VCF file in the subprocess. The VCF and BAM need to be indexed for that, but this requirement would be worth it. Some complications arise because one would need to merge the resulting VCFs in the correct order into the final output file, but also one could skip pickling and unpickling all the data (it would be enough to just send a chromosome name to the subprocess). I’m not saying you need to rewrite this or that it’s necessarily better, but that’s probably what I would recommend if someone would want to parallelize `whatshap phase`.

This sounds like a big restructuring in the code. I think it would help a lot, if the alignment processing (i.e. determining alleles and realignment) was parallelized. An indexed BAM filed could be divided into chunks, which are then processed in parallel. I had a look into this a while ago, but it was not trivial. I think the problem was, that when I invoked the fork via the multiprocessing functions, I got complains that some pysam classes could not be pickled. Probably this can be fixed by restructuring the code, such that the BAM readers from pysam are created inside the worker threads, not before. But I did not follow this, as I did not want make the PR too big.
Another possibility to make the alignment processing faster would be to port it to C++. However, this would require a lot of code to be ported with uncertain success. I think parallel allele detection and realignment is the better approach.

> This should be solvable by using `resource.RUSAGE_BOTH` instead of `resource.RUSAGE_SELF` in the `log_memory_usage` function at
> 
> https://github.com/whatshap/whatshap/blob/491ec8e4b3bfc5624869dd7c09524769fc3b8dcd/whatshap/cli/__init__.py#L193
> 
> . The constant isn’t available on all platforms, though, so perhaps that doesn’t always work.

Thanks for the hint. I will include this.

> To be precise: It’s possible, but probably not much faster than the unparallelized version. I would expect some speedups, though, because one can release the GIL when calling into the C++ code. When the GIL is released, threads can run in parallel even in Python.

Initially I used thread-level parallelism for this, but the speedup was more or less negligble, even though the majority of the phasing time is spent in C++ code. So the GIL was not released automatically. But I was (and mostly still am) completely new to Python multiprocessing. The multi process parallelization seems to work well, though. I don't think it will get faster or need much less memory with threads instead of processes.

marcelm:
> Initially I used thread-level parallelism for this, but the speedup was more or less negligble, even though the majority of the phasing time is spent in C++ code. So the GIL was not released automatically.

Yepp, you would need to place the long-running C++ function calls in a `with nogil:` section. But that’s just for future reference, using multiprocessing is fine and probably the better way since then even the sections where you do need the GIL can be parallelized.

schrins:
The memory consumption is now logged properly. For the run above I acutally got 2.247 GB. Should anything else be done for this PR?

whatshap/cli/__init__.py
```diff
@@ -188,7 +188,13 @@ def read(self, chromosome, variants, sample, *, read_vcf=True, regions=None):
         return readset, vcf_source_ids
 
 
-def log_memory_usage():
+def log_memory_usage(multiprocessing=False):
```
marcelm, line 191: That parameter name seems a bit overspecific. Can you perhaps call it `include_children`?

whatshap/cli/polyphase.py
```diff
@@ -992,12 +992,13 @@ def compute_linkage_based_block_starts(readset, pos_index, ploidy, single_linkag
             continue
         q = Queue()
         q.put(i)
+        merged_clust[i] = new_num_clust
```
marcelm, line 995: In general, it would be best to have separate commits for unrelated changes, but keep it as it is now.

marcelm (changes requested):

marcelm (approved):

PR #282, opened 2021-02-01
Change u_int64_t to the more standard uint64_t

iansealy:
So WhatsHap compiles on Alpine Linux.


marcelm:
Thanks! Is that all that’s needed for Alpine?

PR #296, opened 2021-03-27
add whatshapdenovo

HaploKit:
This is a new method for haplotype aware de novo assembly of diploid genome from long reads, which is based on 'whatshap'.


tobiasmarschall:
Hi Xiao,
[cc @ptrebert @marcelm]

thanks for developing this, much appreciated. I assume this pull request is in response to an email exchange I had with Alex. From a brief look at the code, this looks like a pipeline (rather than, e.g. a subcommand). It also comes with a number of dependencies to tools that we'd want to avoid. I think we'd need to discuss if/how this can be integrated in the whatshap code base and whether that makes sense. Also, I'd be curious to understand the method better and how it relates to previous approaches (e.g. the pipeline Peter implemented here: https://doi.org/10.1038/s41587-020-0719-5). 

Best,
Tobias
PS: I wanted to "CC" Alex, but I don't know his github user. Add him if you have it....

xiaoluo91:
> Hi Xiao,
> [cc @ptrebert @marcelm]
> 
> thanks for developing this, much appreciated. I assume this pull request is in response to an email exchange I had with Alex. From a brief look at the code, this looks like a pipeline (rather than, e.g. a subcommand). It also comes with a number of dependencies to tools that we'd want to avoid. I think we'd need to discuss if/how this can be integrated in the whatshap code base and whether that makes sense. Also, I'd be curious to understand the method better and how it relates to previous approaches (e.g. the pipeline Peter implemented here: https://doi.org/10.1038/s41587-020-0719-5).
> 
> Best,
> Tobias
> PS: I wanted to "CC" Alex, but I don't know his github user. Add him if you have it....

Hi Tobias,
Many thanks for your quick reply. Yes, this is what Alex suggested. I guess he will send you an email. Then we could have a discussion. By the way, Alex's github user is 'aschoen'.

Ijjjjk:
> This is a new method for haplotype aware de novo assembly of diploid genome from long reads, which is based on 'whatshap'.

PR #302, opened 2021-04-08
 added fasta and gfa formats, and splitting by cluster.

maryamghr:


whatshap/cli/split.py
```diff
@@ -48,12 +50,15 @@ def add_arguments(parser):
...
+    arg('--split-by-cluster', default=False, action='store_true',
+        help='Split by given (chromosome) clusters.')
     arg('reads_file', metavar='READS', help='Input FASTQ/BAM file with reads (FASTQ can be gzipped)')
     arg('list_file', metavar='LIST',
-        help='Tab-separated list with (at least) two columns <readname> and <haplotype> (can be gzipped). '
-        'Currently, the two haplotypes have to be named H1 and H2 (or none). Alternatively, the '
-        'output of the "haplotag" command can be used (4 columns), and this is required for using '
-        'the "--only-largest-block" option (need phaseset and chromosome info).')
+        help='Tab-separated list with (at least) two columns <readname> and <haplotype>, or at least '
+            'three columns <readname>, <chromosome-direction cluster>, and <chromosome cluster> (can be gzipped). '
```
ptrebert, line 58: I assume you think about "Watson/Crick" directionality here, but that's more guessing, i.e. I think the name "chromosome-direction cluster" won't be understood by many users. Could this be renamed - or described - to make clear what the meaning is?

whatshap/cli/split.py
```diff
@@ -172,14 +186,22 @@ def process_haplotag_list_file(
...
+
+    return readname_to_haplotype, known_reads, clusters
 
 
 def _two_column_parser(line):
     cols = line.strip().split("\t")[:2]
     return cols[0], cols[1], None, None
 
 
+def _three_column_parser(line):
```
ptrebert, line 200: why is there a need for a 3-column parser if the third column is always set to None?

whatshap/cli/split.py
```diff
@@ -224,22 +246,45 @@ def _fastq_binary_iterator(fastq_file):
         yield record.name, len(record.sequence), (str(record) + "\n").encode("utf-8")
 
 
-def check_haplotag_list_information(haplotag_list, exit_stack):
+def _gfa_file_iterator(gfa_file):
```
ptrebert, line 249: this looks like it will break easily. afaik, the GFA format specifies the following line types:
comment/header/segment/link/containment/path - all of those should be handled explicitly, and any other non-empty line should raise an explicit error to make debugging easier.
Also, GFA files do not need to contain sequence information, so this way of computing the node length may not give the desired result.

whatshap/cli/split.py
```diff
@@ -254,24 +299,32 @@ def check_haplotag_list_information(haplotag_list, exit_stack):
...
     :param output_h2:
     :param output_untagged:
+    :param output_folder:
+    :param split_by:
+    :param clusters:
     :param exit_stack:
     :return:
     """
     potential_fastq_extensions = ["fastq", "fastq.gz", "fastq.gzip" "fq", "fq.gz" "fq.gzip"]
+    potential_fasta_extensions = [".fasta", ".fa"]
```
ptrebert, line 318: why no gzip support?

whatshap/cli/split.py
```diff
@@ -419,19 +485,32 @@ def run_split(
         timers.start("split-iter-input")
 
         for read_name, read_length, record in input_iterator(input_reader):
-            read_counter["total_reads"] += 1
-            if discard_unknown_reads and read_name not in known_reads:
-                read_counter["unknown_reads"] += 1
+            if type(read_name) != list:
```
ptrebert, line 488:
```suggestion
            if isinstance(read_name, list):
```

whatshap/cli/split.py
maryamghr: Yes that was what I meant. This is the more exact explanations with the input requirements. Maybe we can add these lines? 
'chromosome/direction clusters indicate the original chromosome and forward/backward direction of reads.'
'each of these cluster columns should be a set of integers in the range [1:n], where n is the number of clusters.' 
'For example in female human genome, the range of second and third columns are [1:46] and [1:23], respectively.'

whatshap/cli/split.py
maryamghr:
```suggestion
    potential_fasta_extensions = ["fasta", "fasta.gz", "fasta.gzip", "fa", "fa.gz" "fa.gzip"]
```

whatshap/cli/split.py
maryamghr: I added all possible gfa line records to the function.

whatshap/cli/split.py
```diff
@@ -172,14 +186,22 @@ def process_haplotag_list_file(
             "reads not overlapping largest phased blocks: {}".format(num_removed_reads)
         )
 
-    return readname_to_haplotype, known_reads
+    clusters = list(clusters)
```
ptrebert, line 192:
```suggestion
    clusters = sorted(clusters)
    # remove following clusters.sort()
```

whatshap/cli/split.py
ptrebert: That sounds much better, yet I think we need to clarify that Strand-seq is used to generate this input (otherwise, I fear that there will be many support tickets asking how to generate the input using Illumina WGS)

whatshap/cli/split.py
ptrebert: I realize that it would be "good practice" to add a small test file and write a test for the GFA parsing (at least I don't see that you changed any tests with this pull request?)

PR #316, opened 2021-06-23
Remove duplicate lines introduced in 70da82

pontushojer:
Minor edit


marcelm:
Good catch! I didn’t look too closely, does this change behavior? Then it should be mentioned in the changelog.

marcelm:
I’ll merge this now and add a changelog entry later if necessary. Thanks!

pontushojer:
> Good catch! I didn’t look too closely, does this change behavior? Then it should be mentioned in the changelog.

I saw no change in behaviour. The `bx_tag_to_readlist` is only used to generate the set `reads_to_consider`, any duplicates introduced in the previous version would just be merged. I also run corresponding pytest for haplotag without errors. Then I guess the changlog does not need updating.

PR #335, opened 2021-10-12
Add option --ignore-sample-name to compare.py 

### LINKED ISSUE #334: Feature request: add flag to ignore sample name when comparing single sample VCFs
Sometimes different names are used to describe the same sample (e.g. HG002 vs NA24385, HG001 vs NA12878), which requires one VCF to be reheadered before comparing.  It'd be nice to have a flag to override this behavior for single-sample VCFs when using `whatshap compare`.

pontushojer:
Fix #334.

To make this work I had to change sample name handling to use a list contain the sample name for each input file.

### DIFF
--- a/whatshap/cli/compare.py
+++ b/whatshap/cli/compare.py
@@ -26,6 +26,8 @@ def add_arguments(parser):
         'to process. If not given, use first sample found in VCF.')
     add('--names', metavar='NAMES', default=None, help='Comma-separated list '
         'of data set names to be used in the report (in same order as VCFs).')
+    add('--ignore-sample-name', default=False, action='store_true', help='For single '
+        'sample VCFs, ignore sample name and assume all samples are the same.')
     add('--tsv-pairwise', metavar='TSVPAIRWISE', default=None, help='Filename to write '
         'comparison results from pair-wise comparison to (tab-separated).')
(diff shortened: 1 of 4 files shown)


whatshap/cli/compare.py
```diff
@@ -995,15 +1022,24 @@ def get_variant_tables(
...
     for vcf_reader in vcf_readers:
         if sample_intersection is None:
             sample_intersection = set(vcf_reader.samples)
         else:
             sample_intersection.intersection_update(vcf_reader.samples)
-        all_samples.update(vcf_reader.samples)
+
+        if ignore_name and len(vcf_reader.samples) > 1:
+            raise CommandLineError(
+                "File {!r} contains multiple samples, option --ignore-sample-name not available.".format(
```
marcelm, line 1038: I’ve stopped using `!r` in cases like these when I want to get quotation marks. It’ll work here at the moment (I think), but `VcfReader.path` can also be a `pathlib.Path`, and then this displays as `File PosixPath('name.vcf') contains multiple samples`.

pontushojer: You are absolutely correct. I tested and it did print `File PosixPath('name.vcf') contains multiple samples` for `pathlib.Path` input. Fixed in https://github.com/whatshap/whatshap/pull/335/commits/2ace49ad2cd9794428accac6ec25bb4d4809a79b.

PR #352, opened 2022-02-01
Use threads option of pysam for haplotag output

cjw85:
`whatshap haplotag` is used in a number of pipelines including [clair3](https://github.com/HKU-BAL/Clair3)'s variant calling. In these pipelines it can be a bottleneck since a) its a fairly slow serial step b) it usually occurs in a synchronisation step so no other tasks can be accomplised until the tagging has finished.

A quick profiling reveals 25% of time is spent on the line:
```
for alignment in bam_reader.fetch(contig=chrom, start=start, stop=end):
```
and 50% of the time is spent on the line:
```
bam_writer.write(alignment)
```
with the rest of the time being spent on more productive matters.

This PR alleviates some of the bottleneck that `whatshap haplotag` can cause by exposing the threads option of the `pysam.AlignmentFile` class to allow use of an htslib threadpool for output compression. Setting this parameter to 4-8 threads I have observed a 2-3.5X improvement in the wall time execution speed of `whatshap haplotag`.

The program is also now at the point where the loading of variants and preparation of haplotag information is taking as long as the read tagging loop. The bulk (88%) of `prepare_haplotag_information` is spent on the line:
```
read_set, _ = phased_input_reader.read(
    variant_table.chromosome, variants, sample, regions=regions
)
```

### DIFF
--- a/whatshap/cli/haplotag.py
+++ b/whatshap/cli/haplotag.py
@@ -59,6 +59,8 @@ def add_arguments(parser):
         'haplotype the primary alignment has been assigned to (default: only tag primary alignments).')
     arg('--skip-missing-contigs', default=False, action='store_true',
         help='Skip reads that map to a contig that does not exist in the VCF')
+    arg('--out-threads', default=1, type=int,
+        help='Number of threads to use for output writing (passed to pysam).')
     arg('variant_file', metavar='VCF', help='VCF file with phased variants (must be gzip-compressed and indexed)')
     arg('alignment_file', metavar='ALIGNMENTS',
         help='File (BAM/CRAM) with read alignments to be tagged by haplotype')
@@ -345,7 +347,7 @@ def compute_shared_samples(bam_reader, ignore_read_groups, vcf_samples):
     return shared_samples
 
 
-def open_output_alignment_file(aln_output, reference, vcf_md5, bam_header):
+def open_output_alignment_file(aln_output, reference, vcf_md5, bam_header, threads=1):
     """
     :param aln_output:
     :param reference:
@@ -379,7 +381,7 @@ def open_output_alignment_file(aln_output, reference, vcf_md5, bam_header):
         kwargs = dict(mode="wc", reference_filename=reference)
     else:
         # Write BAM
-        kwargs = dict(mode="wb")
+        kwargs = dict(mode="wb", threads=threads)
     try:
         bam_writer = pysam.AlignmentFile(
             aln_output, header=pysam.AlignmentHeader.from_dict(bam_header), **kwargs
@@ -453,6 +455,7 @@ def run_haplotag(
     haplotag_list=None,
     tag_supplementary=False,
     skip_missing_contigs=False,
+    out_threads=1
 ):
 
     timers = StageTimer()
@@ -498,7 +501,7 @@ def run_haplotag(
 
         bam_writer = stack.enter_context(
             open_output_alignment_file(
-                output, reference, md5_of(variant_file), bam_reader.header.to_dict()
+                output, reference, md5_of(variant_file), bam_reader.header.to_dict(), threads=out_threads
             )
         )
         haplotag_writer = stack.enter_context(open_haplotag_writer(haplotag_list))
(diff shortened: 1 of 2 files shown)


marcelm:
Cool! This is quite impressive for such a relatively small change. I’ll have a closer look within the next couple of days.

One thought that comes to mind is that we may want to use a higher default than 1 for the number of threads.

cjw85:
I've added an extra commit here after realising why an earlier experiment failed.

Trying to improve performance with:

```whatshap haplotag ... | samtools view -@3 --fast ...```

initially did not achieve a useful improvement. But I've realised that for this to work we need to tell pysam to not compress its output. The second commit here not opens the BAM output with `wb0` if outputting to `sys.stdout`.

This method appears optimal (I needed to set threads=8 with pysam to get similar performance, I cannot explain this) but does require a user to have `samtools`, so the self-contained threaded pysam is still useful.

File: whatshap/cli/haplotag.py [hunk 1]
```diff
@@ -378,8 +381,11 @@ def open_output_alignment_file(aln_output, reference, vcf_md5, bam_header):
             )
         kwargs = dict(mode="wc", reference_filename=reference)
     else:
-        # Write BAM
-        kwargs = dict(mode="wb")
+        # Write BAM, disable compression for stdout
+        if aln_output is sys.stdout:
+            kwargs = dict(mode="wb0", threads=threads)
```
marcelm, line 386: Out of curiosity: How did you find which mode string to use? Mode `wb0` is listed in the pysam documentation, but it’s not documented what it does.

whatshap/cli/haplotag.py
```diff
@@ -378,8 +381,11 @@ def open_output_alignment_file(aln_output, reference, vcf_md5, bam_header):
             )
         kwargs = dict(mode="wc", reference_filename=reference)
     else:
-        # Write BAM
-        kwargs = dict(mode="wb")
+        # Write BAM, disable compression for stdout
+        if aln_output is sys.stdout:
```
marcelm, line 385: I think compression should be only be disabled if stdout is not a regular file. Otherwise, running `whatshap haplotag ... > output.bam` will result in huge files.

Can you do this instead:
```suggestion
        if aln_output is sys.stdout and not stdout_is_regular_file():
```
And then add this somewhere (plus imports):
```
def stdout_is_regular_file() -> bool:
    mode = os.fstat(sys.stdout.buffer.fileno()).st_mode
    return stat.S_ISREG(mode)
```

cjw85: I thought about this. To be more in line with samtools itself I thought of adding an explicit `-u / --uncompressed` option, but thought that a bit clumsy because users would need to know to use the option to yield any benefit from piping (in my use case). Your idea is nicer because it will result in always doing a sensible thing.

I'll implement that.

whatshap/cli/haplotag.py
cjw85: From knowledge of what htslib does with these modes (look through the samtools view source for example). The number is the compression level.

whatshap/cli/haplotag.py
```diff
@@ -59,6 +59,9 @@ def add_arguments(parser):
         'haplotype the primary alignment has been assigned to (default: only tag primary alignments).')
     arg('--skip-missing-contigs', default=False, action='store_true',
         help='Skip reads that map to a contig that does not exist in the VCF')
+    arg('--out-threads', default=1, type=int,
+        help='Number of threads to use for output file writing (passed to pysam). '
+        'For optimal performance, instead write output to stdout and use `samtools view` to compress.')
```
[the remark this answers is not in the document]
marcelm: I mean: instead of backticks

whatshap/cli/haplotag.py
cjw85: Implemented in [3686cf5](https://github.com/whatshap/whatshap/pull/352/commits/3686cf567ce57b9e1a4744f79968fcfb26cf5464).

whatshap/cli/haplotag.py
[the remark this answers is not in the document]
cjw85: Changed in [3686cf5](https://github.com/whatshap/whatshap/pull/352/commits/3686cf567ce57b9e1a4744f79968fcfb26cf5464), I don't know what you opinion is on double vs. single quotes 😛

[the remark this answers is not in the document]
marcelm: Yeah, "regular" was not specific enough ... Double is fine, seems more normal in user-visible text.

marcelm:
Thanks, this is great! I’ll add a changelog entry after merging.

PR #362, opened 2022-03-27
Bayesian polyphase

schrins:
The purpose of this pull request is to merge the `bayesian-polyphase` branch back to the main. It contains very broad changes to the polyphase algorithm. The overall goals were:

1. Improve the compatibility with different data sets.
2. Provide a formally more clear model

While the original algorithm was tested on artificial tetraploid data based on human samples (for the sake of having a ground truth), it turned out to struggle with differently shaped data sets. The main issue seemed to be assumptions that were true for the test data, but not in general. The clustering stage (stage 1) contained a reclustering after the first one, where clusters with too much coverage (i.e. reads from more than one haplotypes are likely to be contained) were split based on SNP sites inside the cluster with enough disagreement. While this worked for the original data set, real plant data with higher SNP density saw on oversplitting of clusters. The threading stage (stage 2) of the algorithm was completely relying on the input genotypes, which sometimes lead to haplotypes completely break apart when the genotype did not match the allele distribution among the reads well enough. The new algorithm now contains an explicit third stage (reordering stage), where the above problems are handled, instead of doing this during the first two stages. Therefore the previous `threading.py` was split into `threading.py` and `reorder.py`. Some shared functionality of the polyphase files was outsourced to `polyphaseutil.py`.

Since the old algorithm was designed with the tetraploid human data in mind, it still performs better on this particular data than the new one. On the other hand, the test data provided by the publication of Shaw et al. [1] took extremely long for the old algorithm and did not produce a meaningful output. The new algorithm is still not flawless on this data (in fact, the clustering almost perfectly resolves the phasing, but some errors are introduced in the threading stage because of strong coverage fluctuation), but at least reasonable. The performance on Hifi-CCS-reads which we used in a different project (located in branch `genetic-polyphase`, pending paper submission) also greatly increased for the new version. This implies that the compatibility towards different data was improved. I am not aware of other regressions compared to the main branch version.

As of now, I have no further additions to this branch. If I should give more details on the changes or there is critique on the current shape of this branch, please let me know.

[1] https://www.researchgate.net/publication/357919425_flopp_Extremely_Fast_Long-Read_Polyploid_Haplotype_Phasing_by_Uniform_Tree_Partitioning


schrins:
Since the extensive code changes in the polyphase files/classes make a code review quite exhausting and since my agenda is to immediately integrate the changes into the `genetic-polyphsae` branch and continue working on this module, would it be an option to leave everything related to the polyphase module for now and just review whether the other changes are of any concern for the other modules? That would leave the following files:

- src/binomial.h
- src/binomial.cpp
- src/genotype.h
- src/genotype.cpp
- whatshap/cpp.pxd (At some point all tabs got replaced by 4 spaces, I think some code linter might have done this, but the current version of black does not seem to complain)

marcelm:
Hi Sven, sorry I missed the PR – you opened it while I was on vacation and then it got buried among all the other notifications from this repository. I assume your latest comment is directed at me since we probably don’t have any else who could review this PR.

Looking at the files you listed, these look ok to me. I definitely cannot review the rest, but that is fine.

Generally, as you probably know yourself, smaller, more focused PRs are preferred. In this case, it would probably not have changed much, though, because I wouldn’t have been able to comment so much on the polyphase code anyway.

I have two suggestions.

First, please add an entry to `CHANGES.rst`.

Second:
> Therefore the previous `threading.py` was split into `threading.py` and `reorder.py`. Some shared functionality of the polyphase files was outsourced to `polyphaseutil.py`.

Have you considered moving all the polyphase-related modules into their own `polyphase/` directory? So move `threading.py` to `polyphase/threading.py` etc, and move `polyphaseutil.py` to either `polyphase/util.py` or `polyphase/__init__.py`. I don’t know though whether that can be done cleanly and whether the `.pyx` files would also need to be moved.

Feel free to merge this yourself when you have added at least the changelog entry. Reorganizing the modules can also be done in a different PR.

schrins:
Hi Marcel, thanks for your reply. I added a new entry in `CHANGES.rst` mentioning this pull request. I tried to move the polyphase related files to a subfolder in the `whatshap`-directory, but for the `.pyx` this does seem to work so easily, because it references the `cpp.pxd` and I did not figure out yet how to do this, if this file is one directory level above. Maybe we postpone the reorganization of the files to a seperate PR.

The CI test now suddenly fails for `black` even though I just edited the changelog file and the tool also does not work locally anymore for me in version 22.1. It seems like there is an issue with another package and the developers propose to update to 22.3: https://github.com/psf/black/issues/2964
The linter passes for me if I use 22.3 locally.

CHANGES.rst
```diff
@@ -2,6 +2,13 @@
 Changes
 =======
 
+v1.4 (2022-04-07)
```
marcelm, line 5: So I guess that means you want me to release v1.4 today ...

I usually just write `development version` as the title, with neither date nor version number, and then update it just before making the release.

(But no problem, I can make a release today.)

marcelm:
> Maybe we postpone the reorganization of the files to a seperate PR.

I agree. If it’s not straightforward, it’s better to do it separately. 

> The CI test now suddenly fails for `black` even though I just edited the changelog file and the tool also does not work locally anymore for me in version 22.1. It seems like there is an issue with another package and the developers propose to update to 22.3: [psf/black#2964](https://github.com/psf/black/issues/2964) The linter passes for me if I use 22.3 locally.

Yes, I noticed the problem as well and have fixed it already d2bf1304b38ba2280012bcbc65acc0572f3dacd3. Because all the other tests pass, I’ll merge this now.

PR #377, opened 2022-05-31
Cluster editing speedup

schrins:
I found a nice shortcut in the cluster editing heuristic, which does not change the results (or at least it should not and I did not find any differences so far), but makes the heuristic run in less than a third of the old time. This makes the regular polyphase algorithm 20-30% faster. The upcoming genetic polyphase should profit much more, because the cluster editing is much more of a bottleneck there, at least on large instances. I also added/updates the testcases for cluster editing.

I did not edit the `CHANGES.rst` yet, because I first wanted to wait for approval.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -14,6 +14,8 @@ development version
   back to the less accurate algorithm.
 * :issue:`371`: ``whatshap split`` crashed when attempting to split
   reads in a FASTQ file by haplotype.
+* :pr:`377`: Speed-up of about 20-30% for ``whatshap polyphase`` via
+  some optimizations in the read clustering algorithm.
 
 v1.4 (2022-04-07)
 -----------------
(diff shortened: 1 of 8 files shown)


marcelm:
Awesome! Do you have anyone in particular in mind who can approve this? I’m not familiar with the code, so I cannot review it. I think if the tests pass, it should go in after adding a changelog entry as you said.

schrins:
There is probably no one besides me who knows the cluster editing C++ code. I added a few more changes, most of them actually were in the main branch for some reason (or I should rather say: they were not in the feature branch for some reason) and tested some instances again to make sure the output phasings are identical. If there are no objections, I think we should merge this branch.

Speed-ups are hard to measure (see below), so I stayed with a conservative 20-30% speed-up in the changelog.
For a sample of the tetraploid human data from the polyphase paper, the cluster editing time went down from 72 to 29 seconds and the total time went down from 198 to 155 seconds. The extrem case is a sample of the test data provided by another polyphase algorithm (Flopp by Shaw et al. 2022), because the reads form extremely big clusters: The clustering time went down from 63 to 2 (!) seconds with overall time going down from 114 to 52 seconds.

PR #379, opened 2022-06-15
Genetic polyphase

schrins:
I do have another pull request, this time regarding a project about polyploid phasing with pedigree information. The method itself is published here: https://www.sciencedirect.com/science/article/pii/S2589004222007325

As for the implementation, the new algorithm has been placed in a new module named `polyphasegenetic`. This module was created as a preliminary solution to not disturb the code of the normal `polyphase` module. There are plans to integrate both algorithms into a hybrid one, because it would be more powerful to not only phase based on genotypes (what `polyphasegenetic` currently does), but also add read information as well. However, we are still working on a model/method to properly do this, so this will not be finished in the very near future. If it is not desirable to introduce and later retract a module, I can try to merge it into the existing module (though this will of course make the diff bigger).

Besides from the new block of code files, there are some changes in the existing `vcf.py`. The background is, that `polyphasegenetic` needs two new things out of VCF files:

1. Per-variant, per-sample allele depths (in an `AD` field in the format `<count of allele 0>,<count of allele 1>, ..., <count of allele_m>`).
2. The ability to read multi-allelic variants into a variant table.

While the algorithm actually does not process variants with more than two different alleles among the phased samples, there are quite some variants, where the alleles are named with `2` or higher, but there are still only two different alleles present in the samples (e.g. the reference allele does not occur in the samples). I made the existing `VcfVariant` abstract and created two new subclasses `VcfBiallelicVariant` and `VcfMultiallelicVariant`, where the first one has the same functionality than the old class. The latter can additionally return a list of alternative alleles and comparison and normalization are bit more complex. I replaced all explicit occurences of the old class with the new `VcfBiallelicVariant` and the test cases seem to work. Reading these two new kinds of information is controlled via flags which default to False. Please note that this does not make whatshap MAV-capable as a whole, because allele detection in reads still does not support it (although I would like to add this to the `polyphase` module at some time).

On the project side, the new module adds a dependency, namely the (free) ILP solver PuLP. It is used to solve a scheduling problem in one of the algorithm's steps. PuLP is available on conda and pip.

I added a bunch of test cases for the module, although there are currently no tests to run the entire module, since I will work on merging it into the `polyphase` module anyways.

P.S.: What I would request a review for are the changes in `vcf.py`, because they potentially affect other modules. And second, the new dependency of PuLP.


schrins:
Any thoughts about my "P.S." above @marcelm ?

marcelm (commented):
I’ve reviewed as much as I had time for today, but I’m not quite finished, yet. I looked mostly into stylistic issues; I still need to think about whether splitting up VcfVariant like this is ok.

schrins:
Regarding the split VcfVariant classes: In principle, the multi-allelic version can do everything what the bi-allelic version does. The only downside I see is the increased memory footprint for one-element-lists compared to a single value. If that is not an issue for the existing modules, one could make just one class that supports everything. The function `get_alt_allele` is obviously the same then as `get_alt_allele_list(...)[0]` and could be removed. This, however, would induce refactoring across the entire project.

marcelm:
Ok, then I think splitting VcfVariant is ok as an intermediate step and then we can merge them again later.

whatshap/offspringscoring.py
```diff
@@ -0,0 +1,279 @@
...
+from scipy.special import binom as binom_coeff
+
+from whatshap.core import TriangleSparseMatrix, ProgenyGenotypeLikelihoods
+from whatshap.vcf import VariantTable
+from whatshap.variantselection import VariantInfo
+
+logger = logging.getLogger(__name__)
+
+
+class CachedBinomialCalculator:
```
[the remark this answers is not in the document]
schrins: Seems to work fine and to be even slightly faster than the wrapper class.

whatshap/offspringscoring.py
```diff
@@ -0,0 +1,279 @@
+import logging
```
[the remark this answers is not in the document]
schrins: What do you mean with module-level? Once per module (like `phase`, `polyphase`, etc.) or once per (reasonably big) file?

schrins:
Finally had time to go through all the suggested changes. Only remaining question is the one about the docstring.

schrins:
One more thought about MultiallelicVcfVariant: When initializing a one-element list, it has an overhead of 64 Bytes (as opposed to 88 Bytes when starting with an empty list and appending, probably because it reserves more slots when growing incrementally). The two string for ref and alt consume 48 Bytes each as overhead, plus the length of the strings themselves. The position uses 28 Bytes, because Python stores integer as objects. And the variant objects itself consumes another 56 Bytes. The worst case for a SNP would then be:
56 + 28 + 2*(48+1) = 182 Bytes for old VcfVariant
56 + 28 + 2*(48+1) + 64 = 246 Bytes for MultiallelicVcfVariant
This makes an additional 1GB memory consumption per 16,7 million (!) variants in the table, regardless of sample size. For human data I would assume one million variants is already plenty for a single chromosome, so the efficiency regression for diploid phasing should be moderate. In general, the relative increase is not extreme. If memory is the only concern for a combined implementation, we could as well resolve it in this PR.

whatshap/offspringscoring.py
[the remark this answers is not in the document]
marcelm: I meant a triple-quoted string at the very beginning of the file that explains why the file exists, somewhat like this one:

https://github.com/whatshap/whatshap/blob/43af40d20eb66f0b8bc3018fb786cf6fe3e0f6cc/whatshap/pedigree.py#L1-L3

This isn’t such a good example (because it essentially just restates the file name), but perhaps one sentence about what this offspring scoring is or does would be nice.

marcelm:
I recently pushed some commits that conflict with this PR. I can try to fix the merge conflicts if that’s ok with you.

Regarding memory usage:
> In general, the relative increase is not extreme.

I agree. The `VariantTable` is a bit ineffecient memory-wise, so perhaps we should improve that a little bit (possibly by adding yet another Cython module and keeping the data in C structs), but we should do that separately, no need to increase the size of this already large PR even further.

schrins:
Postponing the issue with the variant classes is fine with me. The longer the merge is postponed, the more conflicts will arise over time. Regarding the conflicts with the current head: They seem to come from outsourcing `polyphase_solver.pyx`, which leads to lots of header changes, obviously. I don't mind fixing them, if you did not start already.

marcelm:
If you have the time, please go ahead and fix the merge conflicts, I realized I have less time than I thought.

schrins:
Should be done.

whatshap/polyphaseplots.py
```diff
@@ -476,3 +479,764 @@ def draw_threading(
...
+def draw_phase_comparison(
+    haplotypes,
+    phased_positions,
+    sample_cov,
+    co_parent_cov,
+    progeny_cov,
+    ground_truth_table,
+    path,
+):
+    try:
```
marcelm, line 726: This `try` block is very large. Since you wrote that this code is only for debugging (is it?), I think it’s fine if you just don’t have a try/except at all and instead simply let the ImportError bubble up. A plotting function shouldn’t fail silently when it cannot plot; it should raise an error. If whoever is calling the function wants to ignore a plotting failure, that’s a different thing, but then the ImportError should be caught there.

Same goes for a couple of other `import matplotlib` occurrences below.

whatshap/polyphaseplots.py
```diff
@@ -476,3 +479,764 @@ def draw_threading(
...
+            for pos in range(0, compared_to_phased_pos[-1], 1):
+                x = x_scale * (pos + 1 - 2 * x_margin)
+                axes.text(
+                    x,
+                    ploidy + 2 * y_margin,
+                    phased_positions[pos] + 1,
+                    fontsize=15,
+                    rotation="vertical",
+                )
+        except:
```
marcelm, line 861: That’s a bit broad. What is the actual exception you want to ignore?

whatshap/polyphaseplots.py
```diff
@@ -476,3 +479,764 @@ def draw_threading(
...
+            [
+                b / (a + b) if a + b != 0 else 0
+                for a, b in zip(covered_progeny_non, covered_progeny_err)
+            ],
+            color="tab:green",
+        )
+        plt.title("Ratio percentiles (progeny)")
+        plt.xlabel("Normalized ratio (1 = median ratio)")
+        plt.ylabel("Percentile")
+        savefig(path[:-4] + ".progeny-percentile.pdf", bbox_inches="tight")
```
marcelm, line 1102: You should be able to use `fig.savefig(...)` here, and then you don’t need the `from pylab import savefig` (pylab is more for interactive use AFAIK).

marcelm (commented):
So is this ready to be merged from your side? I browsed through it a little bit and have added a couple of minor comments. I think I’d like to request a docstring for the following new modules before merging:
* `whatshap/clusterarrangement.py`
* `whatshap/offspringscoring.py`
* `whatshap/variantselection.py`

These are new and I have no idea why they were added or what they are for. I’d like to at least be able to tell that they are relevant for genetic polyphasing so that I know I can ignore them when I work on a different subcommand. Just a sentence or so is enough, nothing long.

setup.py
```diff
@@ -102,6 +103,7 @@ def build_extensions(self):
         "biopython>=1.73",  # pyfaidx needs this for reading bgzipped FASTA files
         "scipy",
         "xopen>=1.2.0",
+        "pulp>=2",
```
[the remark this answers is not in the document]
schrins: Yes. Even for whole chromosomes the running time was absolutely fine. If the solver was not sufficient, I would not use an ILP in the first place, because cross-linking another (proprietary) solver is quite a pain and would make the tool too difficult to access for users. Should PuLP ever break, it is possible to solve the underlying problem with some flow-based graph algorithm in polynomial time, but it was much easier to set up a small ILP than diving into some flow implementations.

whatshap/polyphaseplots.py
schrins: I removed the try-catch for these cases. I think I added them to not just pass the exception upwards, but since this module is not intended to be called by users anyways, this is not necessary. In general, the import instructions are inside the methods, because I import this module in other places (e.g. cli/polyphase.py) and putting the import at the top would add a matplotlib dependency to whatshap just for the sake of some debugging methods.

whatshap/polyphaseplots.py
schrins: I don't remember. I removed the try-catch and everything seems to work fine.

whatshap/polyphaseplots.py
marcelm: Yes, I’d also prefer not having the matplotlib dependency, so importing it within the function is a good way to avoid this.

setup.py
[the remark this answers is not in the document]
marcelm: Great, nice to not have to rely on proprietary solvers!

marcelm:
Thanks for the updates. Please merge when you are ready!

schrins:
From my side, this can be merged. There are still things to do on the genetic module, but it is functional now as it is.

marcelm (approved):

PR #384, opened 2022-07-15
whatshap stats - Update NG50 calculation

pontushojer:
Split interleaved phase blocks to maintain as much of the phased variants as possible. This will make NG50 values a bit more fair compared to how they were calculated previously. 

If one phase block is fully encompassed by another block the inner block would determine the end of the outer block on the left side. The right side variants in the outer block were however scrapped in the current version leading to some deflation of the NG50 value. 

I made a small example below to clarify.

**Original phase blocks**
```
Phase block 1:    A--G--G--T----C------C--T--C
Phase block 2:               A----G--C
```

**Previous split**
```
Phase block 1:   A--G--G--T
Phase block 2:              A--G--C
```

**Updated split**
```
Phase block 1.1:  A--G--G--T
Phase block 1.2:                       C--T--C
Phase block 2:               A--G--C
```

### DIFF
--- a/tests/test_run_stats.py
+++ b/tests/test_run_stats.py
@@ -85,3 +85,46 @@ def test_stats2(tmp_path):
     assert entry_all.variant_per_block_sum == "8"
     assert entry_all.bp_per_block_sum == "750"
     assert entry_all.block_n50[:-1] == "350"
+
+
+def test_overlapping_phaseblocks(tmp_path):
+    """
+    We have three phaseblocks on chrA which is 1000 bp
+
+        chrA:100-700 --> 600 bp
+        chrA:410-470 --> 60 bp
+        chrA:800-950 --> 150 bp
+
+    Total block sum should be 600 + 60 + 150 = 810 bp
+
+    For NG50 the first block is split since the second block overlaps, now we have four blocks
+
+        chrA:100-350 --> 250 bp
+        chrA:410-470 --> 60 bp
+        chrA:500-700 --> 200 bp
+        chrA:800-950 --> 150 bp
+
+    Half of the total length is 1000 * 0.5 = 500 bp.
+    Let's calculate NG50 by adding block lengths in descending order until we exceed 500 bp
+
+        block   length  total   >500
+        1       250     250     no
+        2       200     450     no
+        3       150     600     yes ->  NG50 = 150 bp
+    """
+
+    outtsv = tmp_path / "output.tsv"
+    run_stats(
+        vcf="tests/data/phased_overlapping.vcf",
+        tsv=outtsv,
+        sample="sample1",
+    )
+    lines = [l.split("\t") for l in open(outtsv)]
+    assert len(lines) == 2
+    Fields = namedtuple("Fields", [f.strip("#\n") for f in lines[0]])
+    entry = Fields(*lines[1])
+
+    assert entry.chromosome == "chrA"
+    assert entry.blocks == "3"
+    assert entry.bp_per_block_sum == "810"
+    assert entry.block_n50[:-1] == "150"
--- a/tests/test_run_stats.py
+++ b/tests/test_run_stats.py
@@ -13,7 +13,8 @@ def test_stats1(tmp_path):
         sample="sample1",
         chr_lengths="tests/data/chr-lengths.txt",
     )
-    lines = [l.split("\t") for l in open(outtsv)]
+    with open(outtsv) as f:
+        lines = [l.split("\t") for l in f]
     assert len(lines) == 4
     Fields = namedtuple("Fields", [f.strip("#\n") for f in lines[0]])
     entry_chrA, entry_chrB, entry_all = [Fields(*l) for l in lines[1:]]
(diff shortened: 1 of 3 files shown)


marcelm:
Thanks and sorry for the delay – this came in while I was on vacation. I I’ll look into this and your other PRs probably next week.

pontushojer:
> Thanks and sorry for the delay – this came in while I was on vacation. I I’ll look into this and your other PRs probably next week.

Yeah, sorry for the PR spam ;) No rush here! I just had some spare time during summer to fix some things I had been considering for a while.

whatshap/cli/stats.py
```diff
@@ -194,6 +198,46 @@ def add_heterozygous_variants(self, variants: int):
     def add_heterozygous_snvs(self, snvs: int):
         self.heterozygous_snvs += snvs
 
+    def get_nonoverlapping_blocks(self) -> List[PhasedBlock]:
+        """Split phase blocks into nonoverlapping subblocks"""
+        split_blocks = []
```
marcelm, line 203: Please move this line to just above the `while`: This function does two things 1) preprocess blocks, 2) split blocks. Logically, this line belongs to the second part.

whatshap/cli/stats.py
```diff
@@ -194,6 +198,46 @@ def add_heterozygous_variants(self, variants: int):
...
+                    # Update sorting if right-side block is added.
+                    if len(new_block) > 1:
+                        pos_sorted_blocks.append(new_block)
+                        pos_sorted_blocks = sorted(
+                            pos_sorted_blocks,
+                            key=lambda b: (b.chromosome, b.leftmost_variant.position),
+                            reverse=True,
+                        )
+
+                    # Skip the left-side block is it too short after splitting
```
marcelm, line 234: should be "if it is"

whatshap/cli/stats.py
```diff
@@ -194,6 +198,46 @@ def add_heterozygous_variants(self, variants: int):
...
+                next_block_end = next_block.rightmost_variant.position
+
+                # Check if next block overlapps current. If so split the current block.
+                if (block_end > next_block_start) and (block.chromosome == next_block.chromosome):
+                    block, new_block = block.split(next_block_start, next_block_end)
+
+                    # Update sorting if right-side block is added.
+                    if len(new_block) > 1:
+                        pos_sorted_blocks.append(new_block)
+                        pos_sorted_blocks = sorted(
```
marcelm, line 228: I was a bit worried because the entire list is resorted just because of one new element, but it is probably fine because Python’s sort is (as far as I know) fast with nearly sorted lists. (The alternative is to use a heap.)

I also think that one should only work on one chromosome at a time since blocks cannot span chromosomes, but I realize the original code did it like this, so let’s leave it as it is.

whatshap/cli/stats.py
```diff
@@ -194,6 +198,46 @@ def add_heterozygous_variants(self, variants: int):
...
+        split_blocks = []
+        pos_sorted_blocks = sorted(
+            self.blocks, key=lambda b: (b.chromosome, b.leftmost_variant.position), reverse=True
+        )
+
+        # filter out blocks with only one variant
+        pos_sorted_blocks = [b for b in pos_sorted_blocks if len(b) > 1]
+
+        # iterate over blocks and split if overlapping until no blocks remain. non-overlapping
+        # blocks appended to split_blocks
```
marcelm, line 212: I think the second sentence can be removed

whatshap/cli/stats.py
```diff
@@ -64,6 +65,17 @@ def variants(self):
     def count_snvs(self):
         return sum(int(variant.is_snv()) for variant in self.phases)
 
+    def split(self, split_left: int, split_right: int) -> Tuple["PhasedBlock", "PhasedBlock"]:
+        """Split a phaseblock in two based on given positions"""
```
marcelm, line 69: * Should be "split *this* phase block".
* The docstring isn’t detailed enough ("based on" is unclear). Please describe `split_left` and `split_right` (because intuitively I would expect a `split()` function to only have one split position argument).
* Should there be an `assert split_left <= split_right`?

marcelm (commented):
A few more comments this time, but once again, nothing major. This is a good improvement.

tests/test_run_stats.py
```diff
@@ -85,3 +85,46 @@ def test_stats2(tmp_path):
...
+        3       150     600     yes ->  NG50 = 150 bp
+    """
+
+    outtsv = tmp_path / "output.tsv"
+    run_stats(
+        vcf="tests/data/phased_overlapping.vcf",
+        tsv=outtsv,
+        sample="sample1",
+    )
+    lines = [l.split("\t") for l in open(outtsv)]
```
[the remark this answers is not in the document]
pontushojer: Definitly! I just copied this from the tests above, I will fix this

whatshap/cli/stats.py
pontushojer: Agreed! I will update the docs here and a the `assert split_left <= split_right` is probably a good call.

whatshap/cli/stats.py
pontushojer: Yes, that makes more sense

pontushojer:
I added the changes you suggested. Should be good to merge now!

marcelm:
Thanks once again, will merge and probably make a release soon.

PR #385, opened 2022-07-15
whatshap stats - Speed up chromosome lookup 

pontushojer:
I noticed that when using `whatshap stats` with `--chromosomes` the entire VCF is still read, which is very slow. This PR does two things. 

- If the VCF is indexed (.tbi or .csi) --> lookup chromosomes of interest (given by `--chromosomes`) directly
- If the VCF is not indexed --> stop reading if all chromosomes of intest have been seen

### DIFF
--- a/whatshap/cli/stats.py
+++ b/whatshap/cli/stats.py
@@ -460,7 +460,7 @@ def run_stats(
 
             total_stats += stats
 
-            if given_chromosomes and all(c in seen_chromosomes for c in given_chromosomes):
+            if given_chromosomes and set(given_chromosomes) <= seen_chromosomes:
                 break
 
         if chromosome_count > 1:
--- a/whatshap/cli/stats.py
+++ b/whatshap/cli/stats.py
@@ -382,8 +382,10 @@ def run_stats(
         total_stats = PhasingStats()
         chromosome_count = 0
         given_chromosomes = chromosomes
+        seen_chromosomes = set()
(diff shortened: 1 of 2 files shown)


whatshap/cli/stats.py
```diff
@@ -446,6 +460,9 @@ def run_stats(
 
             total_stats += stats
 
+            if given_chromosomes and all(c in seen_chromosomes for c in given_chromosomes):
```
marcelm, line 463:
```suggestion
            if given_chromosomes and set(given_chromosomes) <= seen_chromosomes:
```

whatshap/vcf.py
```diff
@@ -387,6 +387,10 @@ def close(self):
     def path(self) -> str:
         return self._vcf_reader.filename.decode()
 
+    def index_exists(self) -> bool:
+        """ "Check if VCF is indexed (.tbi or .csi)"""
+        return os.path.isfile(f"{self.path}.tbi") or os.path.isfile(f"{self.path}.csi")
```
marcelm, line 392:
```suggestion
        return self._vcf_reader.index is not None
```

marcelm (commented):
Thanks, this is a good improvement. Just two suggestions.

pontushojer:
Great! I added both of your suggestions as they look good. Did not know about the `.index` attribute for pysam.VariantFile but that is very useful.

marcelm:
Thanks once again!

Issue #408, opened 2022-10-31
whatshap haplotag and split for polyploid


fangdm:
Hi, I am using whatshap to analyze polyploid data, and when I run 'whatshap haplotag',  I find that it can't handle polyploid data, there is no "--ploidy" parameter. The results of haplotag are shown as diploids (H1, H2 and none) in resule "haplotypes.tsv".
Any ideas on how to fix for polyploid?

the command as:
whatshap haplotag -o haplotagged.bam --reference ref.fa phased.vcf.gz aln.best.sort.bam --output-threads=5 --ignore-read-groups 
--output-haplotag-list haplotypes.tsv

Thank you.

Whatshap version 1.6

fangdm:
it's previous command as:
whatshap polyphase -o phased.vcf --reference=ref.fa --mapq 20 --ploidy 3 -t 1 --indels --ignore-read-groups --distrust-genotypes  variants.vcf aln.best.sort.bam

"variants.vcf" was based on  FreeBayes with illumina, and "aln.best.sort.bam" from HIFI data mapping.

Thank you.

schrins:
I never participated in the haplotag command, but there was supposed to be a polyploid version of it at some point. I think it's still hidden in the `polyploid-haplotag` branch, which is based on quite an old whatshap version now. I might have a look at this, when I have the time, because it is a useful feature, but I can make no promises.

Adamtaranto:
Polyploid haplotagging would be super useful! 1++ for this feature.

Issue #267, opened 2020-10-22
Polyploid version of whatshap haplotag


rvicedomini:
Hello,
this question might be also related to #214. Is there any plan to introduce the polyploid version of haplotag (and split) in the main development branch of whatshap?

Thanks,
Riccardo

tobiasmarschall:
Hi Riccardo,

absolutely! Thanks for reminding us. Presently everybody is pretty busy with other tasks, so I expect that might not happen immediately, but please do ping us again if it hasn't happened in (say) a months time.

Best,
Tobias

tobiasmarschall:
PS: I assume you did see the [branch polyploid-haplotag](https://github.com/whatshap/whatshap/tree/polyploid-haplotag), didn't you?

rvicedomini:
Thank you very much for the answer!
Yes I'm already considering the polyploid-haplotag branch. I was just thinking it would be great to have polyploid haplotag in the main branch in order to benefit from all recent whatshap improvements (without having multiple versions installed).

Best,
Riccardo

baozg:
Hi, @tobiasmarschall 

Would the `polyploid-haplotag` will merge into the main branch?

Best,
Zhigui

tobiasmarschall:
That's certainly the plan, but I'm not sure what the status is. @eblerjana or @ptrebert, could you comment?

schrins:
Haplotag should now support polyploid input with #410 being merged. The split command is still diploid, as far as I know. Since there is another issue (#408 ) with the same request, I will close this one and refer to the other for further comments.

baozg:
Great! Thanks for your support!

PR #410, opened 2022-11-04
Version of haplotag that also supports higher ploidies

eblerjana:
This branch contains an updated version of the haplotag command that we had implemented in branch "polyploid-haplotag" quite some time ago. It was used in the whatshap polyphase paper for haplotagging polyploids, but has never been merged to master. The version implemented here now extends the original haplotag command so that it can also handle higher ploidies than 2 (while still working the same way for diploid samples). The ploidy can be specified with parameter `--ploidy` (default is 2). This feature has been requested in several issues, such as #267 and #408.


marcelm:
Oh, nice! Please give me a couple of days until I have some time for a review.

whatshap/cli/haplotag.py
```diff
@@ -206,34 +208,41 @@ def prepare_haplotag_information(
...
 
             if len(l) == 0:
                 continue
             if len(l) > 1:
                 n_multiple_phase_sets += 1
-            phaseset, quality = l[0]
+            phaseset, scores = l[0]
+
+            # find best and second best haplotype scores for this phaseset
+            scores_list = [s for s in enumerate(scores)]
```
marcelm, line 228:
```suggestion
            scores_list = list(enumerate(scores))
```

marcelm (commented):
I’m not too familiar with how haplotag works, but have tried to work through it and it seems to look fine. I think haplotag also needs some minor cleanups, but that’s something for later.

Can you address the minor comments I have before merging?

marcelm:
Thanks!

Issue #407, opened 2022-10-25
Some error about "ValueError: math domain error"


fangdm:
Hello, I am running this command:

`whatshap polyphase -o phased.vcf --reference=ref.fa --mapq 20 --indels --ignore-read-groups --ploidy 3 -t 1 variants.vcf aln.best.sort.bam`

the software runs correctly, but for some contigs, it is raising this error:

```
Traceback (most recent call last):
  File "/hwfssz1/ST_EARTH/Reference/ST_AGRIC/USER/wangfang/02.software/miniconda3/envs/whatshap-env/bin/whatshap", line 10, in <module>
    sys.exit(main())
  File "/hwfssz1/ST_EARTH/Reference/ST_AGRIC/USER/wangfang/02.software/miniconda3/envs/whatshap-env/lib/python3.10/site-packages/whatshap/__main__.py", line 64, in main
    module.main(args)
  File "/hwfssz1/ST_EARTH/Reference/ST_AGRIC/USER/wangfang/02.software/miniconda3/envs/whatshap-env/lib/python3.10/site-packages/whatshap/cli/polyphase.py", line 821, in main
    run_polyphase(**vars(args))
  File "/hwfssz1/ST_EARTH/Reference/ST_AGRIC/USER/wangfang/02.software/miniconda3/envs/whatshap-env/lib/python3.10/site-packages/whatshap/cli/polyphase.py", line 295, in run_polyphase
    ) = phase_single_individual(
  File "/hwfssz1/ST_EARTH/Reference/ST_AGRIC/USER/wangfang/02.software/miniconda3/envs/whatshap-env/lib/python3.10/site-packages/whatshap/cli/polyphase.py", line 421, in phase_single_individual
    phase_single_block(block_id, block_readset, gt_slices[block_id], param, timers)
  File "/hwfssz1/ST_EARTH/Reference/ST_AGRIC/USER/wangfang/02.software/miniconda3/envs/whatshap-env/lib/python3.10/site-packages/whatshap/cli/polyphase.py", line 577, in phase_single_block
    paths, haplotypes = run_threading(
  File "/hwfssz1/ST_EARTH/Reference/ST_AGRIC/USER/wangfang/02.software/miniconda3/envs/whatshap-env/lib/python3.10/site-packages/whatshap/threading.py", line 60, in run_threading
    haplotypes = force_genotypes(
[...]

PR #411, opened 2022-11-07
Fixing issue 407

schrins:
In case a new release version is coming up in order to make the polyploid version of haplotag available, I would like this fix to be included as well. I don't know yet whether #407 is actually fixed, but two other bugs are definitely eliminated with relatively small changes. I might update this branch, either to add a few testcases for the previously existing bug or to elaborate further on the error mentioned in #407 if it is still unfixed.


schrins:
I will merge this for now. I think we should with a new release after the polyploid haplotag is merged.

PR #412, opened 2022-11-10
whatshap stats - Calculate all length statistics on split blocks

### LINKED ISSUE #400: strange values of whatshap stats
Good morning, I had a very strange result when computing phasing statistics on a phased vcf file from illumina data.
I'm using WhatsHap v1.4 (installed with conda) and I'm using python Python 3.9.12.
I used the command 
whatshap stats file_phased.vcf --tsv=file_stats
not providing the file with contig lengths as they are included in the vcf header.
The tsv file shows 0 in the column block_n50 so, if I understood correctly, it means that the length of phased block did not reach 50% of contig length (which is totally expectable due to the type of genomic data). However, what is strange is that the value bp_per_block_sum is a lot higher than the total length of the chromosome, for most of them. I don't get what is going on, does anyone have any ideas? It should be the total sum, in bp, of all phased blocks right?
 Thanks a lot, would appreciate your help.

Guido

pontushojer:
fix #400 

Fixes inflated stats for block lengths in `whatshap stats` if there are overlapping blocks, also see  https://github.com/whatshap/whatshap/pull/384. 

I tested on chromosome `SUPER_3` (116801625 bp) in the VCF referenced in issue #400 where there were a lot of overlapping blocks. 
**Before** 
```
      Sum of lengths: 1036875048    bp
 Median block length:        493.00 bp
Average block length:      16018.71 bp
       Longest block:   75095876    bp
      Shortest block:          1    bp
          Block NG50:          0    bp
```

**After**
```
      Sum of lengths: 48207739    bp
 Median block length:      492.00 bp
Average block length:      744.53 bp
       Longest block:    15389    bp
      Shortest block:        1    bp
          Block NG50:        0    bp
```

As you can see there is a drastic reduction in `Sum of lengths`, `Average block length` and `Longest block`. 

This PR also a removes an additional call to `PhasingStats.get` (now renamed `PhasingStats.get_detailed_stats`) when using the flag `--tsv`.

### DIFF
--- a/whatshap/cli/stats.py
+++ b/whatshap/cli/stats.py
@@ -296,7 +296,7 @@ class PhasingStats:
         n_singletons = sum(1 for size in block_sizes if size == 1)
         block_sizes = [size for size in block_sizes if size > 1]
         # Block length stats calculated from split interleaved blocks to avoid inflating values
-        block_lengths = sorted([block.span() for block in self.split_blocks if len(block) > 1])
+        block_lengths = sorted(block.span() for block in self.split_blocks if len(block) > 1)
         phased_snvs = sum(block.count_snvs() for block in self.blocks if len(block) > 1)
         if block_sizes:
             return DetailedStats(
--- a/whatshap/cli/stats.py
+++ b/whatshap/cli/stats.py
@@ -137,6 +137,54 @@ class DetailedStats:
     phased_snvs: int
     block_n50: float
 
+    def print(self, width: int = 21):
+        # Parameters for value formatting
+        max_integer_width = max(len(str(int(value))) for value in vars(self).values())
+        value_width = max(max_integer_width, 8)
+        format_int = f"{value_width}d"
+        format_float = f"{value_width + 3}.2f"
+
+        print("Variants in VCF:".rjust(width), f"{self.variants:{format_int}}")
+        print(
+            "Heterozygous:".rjust(width),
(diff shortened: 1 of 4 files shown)


whatshap/cli/stats.py
```diff
@@ -239,16 +290,15 @@ def get_nonoverlapping_blocks(self) -> List[PhasedBlock]:
...
 
-    def get(self, chr_lengths=None):
+    def get_detailed_stats(self, chr_lengths: Optional[Dict[str, int]] = None) -> DetailedStats:
         """Return DetailedStats"""
         block_sizes = sorted(len(block) for block in self.blocks)
         n_singletons = sum(1 for size in block_sizes if size == 1)
         block_sizes = [size for size in block_sizes if size > 1]
-        block_lengths = sorted(block.span() for block in self.blocks if len(block) > 1)
+        # Block length stats calculated from split interleaved blocks to avoid inflating values
+        block_lengths = sorted([block.span() for block in self.split_blocks if len(block) > 1])
```
marcelm, line 299: Can this remain a generator expression (without the `[` and `]`)?

marcelm (commented):
Looks good, I think this is a nice improvement! Can you address the one comment I had? Then feel free to merge.

marcelm:
I forgot: This also needs a documentation update and an entry in the changelog. In the docs, I think it makes sense to move the sentence about "interleaved blocks are cut" to the paragraph "The numbers in the following columns describe these blocks."

marcelm:
Thanks!

PR #415, opened 2022-11-11
whatshap stats - Refactoring 

pontushojer:
- Reduce complexity by factoring out parts of `run_stats`. 
- Improve legibility and simplified multiple sections of the code. 
- Improved typing in some sections. 
- Added missing printout from DetailedStats `variant_per_block_sum` that was only included in TSV.

### DIFF
--- a/whatshap/cli/stats.py
+++ b/whatshap/cli/stats.py
@@ -368,6 +368,24 @@ def parse_variant_tables(vcf_reader, chromosomes=None):
         yield from vcf_reader
 
 
+def get_chr_lengths(chr_lengths_file, vcf_reader):
+    if chr_lengths_file:
+        chr_lengths = parse_chr_lengths(chr_lengths_file)
+        logger.info("Read length of %d chromosomes from %s", len(chr_lengths), chr_lengths_file)
+    else:
+        chr_lengths = {
+            chrom: contig.length
+            for chrom, contig in vcf_reader.contigs.items()
+            if contig.length is not None
+        }
+        if not chr_lengths:
+            logger.warning(
+                "VCF header does not contain contig lengths, cannot compute NG50. "
+                "Consider using --chr-lengths"
+            )
+    return chr_lengths
+
+
 @dataclasses.dataclass
 class GtfBlock:
     start: int = 0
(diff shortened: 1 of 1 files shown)


whatshap/cli/stats.py
```diff
@@ -341,13 +345,17 @@ def parse_variant_tables(vcf_reader, chromosomes=None):
...
+    Try to map chromosome length to its name. Use chr_lengths_file if it exists.
+    Otherwice look in VCF header for chromosomes.
+    """
     if chr_lengths_file:
         chr_lengths = parse_chr_lengths(chr_lengths_file)
         logger.info("Read length of %d chromosomes from %s", len(chr_lengths), chr_lengths_file)
     else:
         chr_lengths = {
-            chrom: contig.length
+            str(chrom): contig.length
```
marcelm, line 362: Did you add this to appease the type checker? I assume that it complained because the keys in `vcf_reader.contigs.items()` can be `None`. If that is the case, it would be better to just filter these out instead of converting them to `"None"`. (That is, add a second condition to the `if contig.length is not None`.)

whatshap/cli/stats.py
```diff
@@ -341,13 +345,17 @@ def parse_variant_tables(vcf_reader, chromosomes=None):
         yield from vcf_reader
 
 
-def get_chr_lengths(chr_lengths_file, vcf_reader):
+def get_chr_lengths(vcf_reader: VcfReader, chr_lengths_file: Optional[str] = None) -> Dict[str, int]:
+    """
+    Try to map chromosome length to its name. Use chr_lengths_file if it exists.
+    Otherwice look in VCF header for chromosomes.
```
marcelm, line 355: This isn’t very accurate. Suggestion: "Return a dictionary that maps a chromosome name to the chromosome’s length. The mapping is read from `chr_lengths_file` if provided, and from the VCF header otherwise."

marcelm (commented):
Looks really good! Just a couple of minor comments.

whatshap/cli/stats.py
pontushojer: The reason for this was actually that the type for `chrom` was `Union[int, str]` (see https://github.com/pysam-developers/pysam/blob/26f3a7f0fb03cf3347d049f12c8a87ec47acd74e/pysam/libcbcf.pyi#L110). I am not sure why but I agree this maybe not the best solution. After thinking so more I think its better to use the `contig.name` attribute instead of the key since this should be a `str`. Will fix!

whatshap/cli/stats.py
pontushojer: Yeah this is better.

pontushojer:
@marcelm Thanks for the review. Fixed the issue pointed out.

marcelm:
Nice, thanks!

PR #416, opened 2022-11-14
Update docs for whatshap stats to match code

### LINKED ISSUE #414: whatshap stats - definition of phased variants
I am a bit confused on this after looking through the code and documentation. 

This is how the number of phased variants (parameter/column `phased` in output) is currently calculated.
https://github.com/whatshap/whatshap/blob/ee79863c30b59a9031812a99ea80b7666e9342cb/whatshap/cli/stats.py#L304

i.e. its the number of phased variants in non-singleton blocks. This also means that it will be that same values as for `variant_per_block_sum` 

https://github.com/whatshap/whatshap/blob/ee79863c30b59a9031812a99ea80b7666e9342cb/whatshap/cli/stats.py#L312

which makes this statistic a bit redundant.

It makes sense in a way that blocks of length one are not considered phased in these statistics. However in the documentation `phased` is currently defined as: 
> The number of biallelic, heterozygous variants that are marked as phased in the input VCF. This is again a subset of heterozygous_variants. Also, phased + unphased + singletons = heterozygous_variants.

The first sentence would include singletons if they are still marked phased. But then the statment "phased + unphased + singletons = heterozygous_variants" is no longer true. 

I also noted a statement below this stating "singletons + variant_per_block_sum = phased". This is currently the same as "singletons + phased = phased" which is only true if singletons are zero.  

**TLDR** What definition for `phased` (the number of phased variants) should we have in the statistics? I think it should be as now, i.e. excluding singletons, in which case the documentation should be updated to clarify this.

pontushojer:
fix #414

### DIFF
--- a/doc/guide.rst
+++ b/doc/guide.rst
@@ -547,9 +547,9 @@ unphased
     This is also a subset of *heterozygous_variants*.
 
 phased
-    The number of biallelic, heterozygous variants that *are* marked as phased in the input VCF.
-    This is again a subset of *heterozygous_variants*.
-    Also, phased + unphased + singletons = heterozygous_variants.
+    The number of biallelic, heterozygous variants that *are* marked as phased in the input VCF, excluding singletons.
+    This is again a subset of *heterozygous_variants*. Add *singletons* to get the total number of variants marked as phased in the VCF.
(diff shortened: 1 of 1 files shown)


marcelm:
LGTM

PR #419, opened 2022-11-15
whatshap stats - Add percentage of phased heterozygous calls to output

### LINKED ISSUE #143: Add documentation on GTF output for interleaved blocks / Add additional output to stats subcommand
**[Original report](https://bitbucket.org/whatshap/whatshap/issue/143) by madhubioinfo (Bitbucket: [madhubioinfo](https://bitbucket.org/madhubioinfo), GitHub: [madhubioinfo](https://github.com/madhubioinfo)).**

----------------------------------------

For visualizing the phasing results , there is gtf file created, does the 3rd and 4th column represents the start and end co-ordinates of haplotype blocks? 

Question no 2.
The stats is printing the information of how many blocks has been phased, Is it possible to calculate the total % of haplotype phased?

pontushojer:
fix #143 (The issue is also about GTF documentation but that is already fixed)
 
**Example of new output**
Command
```
whatshap stats tests/data/phased3.vcf --chr-length tests/data/chr-lengths.txt --chromosome chrA
```
Output

```
---------------- Chromosome chrA ----------------
      Variants in VCF:        9
         Heterozygous:        9    (       5    SNVs)
               Phased:        4    (       2    SNVs)
  Heterozygous phased:       44.4% (      40.0% SNVs)
             Unphased:        5    (not considered below)
           Singletons:        0    (not considered below)
               Blocks:        1

Block sizes (no. of variants)
         Sum of sizes:        4    variants
    Median block size:        4.00 variants
   Average block size:        4.00 variants
        Largest block:        4    variants
       Smallest block:        4    variants

Block lengths (basepairs)
       Sum of lengths:      350    bp
  Median block length:      350.00 bp
 Average block length:      350.00 bp
        Longest block:      350    bp
       Shortest block:      350    bp
           Block NG50:        0    bp
```


doc/guide.rst
```diff
@@ -555,6 +555,12 @@ phased_snvs
     The number of biallelic, heterozygous SNVs that are marked as phased in the input VCF.
     This is a subset of *phased*.
 
+pct_phased
+    The percentage of heterozygous variants that are phased. Same as *phased* / *heterozygous_variants*.
+
```
marcelm, line 560: You should distinguish between a fraction and a percentage: The percentage is the fraction times 100. The formula here describes the fraction, but the text says percentage.

I haven’t made up my mind whether it makes more sense to report percentages or fractions in general, so I’ll leave it up to you, but it should match the description and be consistent.

For readability, I would like to not abbreviate percent to pct.

Also, for a different project, I read a standard for a tabular file format (reporting V/D/J rearrangements). They say that their column names follow the convention that the [scope narrows from left to right](https://docs.airr-community.org/en/v1.3.1/datarep/rearrangements.html#custom-columns). I think this is a good way to do it, and we somewhat follow this convention already here, so to be consistent, we should call the new columns `phased_percentage` (or `phased_fraction`) and `phased_snvs_percentage` (or `phased_snvs_fraction`).

pontushojer: > You should distinguish between a fraction and a percentage: The percentage is the fraction times 100. The formula here describes the fraction, but the text says percentage.

Yes, this was a bit sloppy by me. 

> I haven’t made up my mind whether it makes more sense to report percentages or fractions in general, so I’ll leave it up to you, but it should match the description and be consistent.

For me reporting fractions in the TSV makes more sense. In the stdout I think it is nicer to show percentage since it is easier to read. 

> Also, for a different project, I read a standard for a tabular file format (reporting V/D/J rearrangements). They say that their column names follow the convention that the [scope narrows from left to right](https://docs.airr-community.org/en/v1.3.1/datarep/rearrangements.html#custom-columns). I think this is a good way to do it, and we somewhat follow this convention already here, so to be consistent, we should call the new columns `phased_percentage` (or `phased_fraction`) and `phased_snvs_percentage` (or `phased_snvs_fraction`).

Thanks for sharing this! I have not thought too deeply about this before but this make more sense indeed.

pontushojer:
Thanks for the review @marcelm! I have now updated so that the TSV outputs the columns `phased_fraction` and `phased_snvs_fraction` and update the docs to match this output.

marcelm:
Sorry, I had forgotten. Looks good now!

Issue #427, opened 2023-01-02
Whatshap polyphase on INDELs not working?


Coryza:
Hi,

I'm running Whatshap polyphase (latest version) on a VCF file generated with Freebayes;
whatshap polyphase genome.vcf genome.bam --ploidy 3 --ignore-read-groups **--indels** --mapping-quality 0 --reference genome.fasta

Interestingly, I observe that no INDELs have been phased. After haplotagging I can manually clearly observe dozens of examples where INDELs are present in 1 haplotype but not in the other. What am I overlooking? 
```

grep -c '^[^#]' genome.vcf
40768
grep -c '|' genome.vcf
7882
grep '|' genome.vcf | awk '{print $4}' | sort | uniq -c
1782 A
2162 C
2197 G
1741 T
```

![polyphase](https://user-images.githubusercontent.com/7194842/210248187-808cf23f-c3db-4531-8177-7b6d6c261299.png)
_INDELs observed in the visualization are present in the VCF, unphased_

Coryza:
Similar issue to #370?

Coryza:
Upon converting MNP to SNP's / simplified INDELs with vcfallelicprimitives all SNPs are properly phased, INDELS (e.g. GA>A) are still not phased.

schrins:
Hi,

I totally overlooked this issue. On first sight everything looks fine from your side. The polyploid phaser is oblivious to the variant type, because it only "sees" which read has which allele (0, 1, ...) for which variant position. This means, there must be an issue with the allele detection before. What happens if you leave out the `--reference` argument? This influences allele detection, because instead of deriving the alleles of each read from the cigar string only, it would additionally do some partial realignments to improve allele detection (hence the need of a reference genome). From experience this does not always lead to better results for non-human reference genomes.

If this does not help, would it be possible to share a small part of the problematic dataset? It would only need to contain one of two indel variants, so I can reproduce the bug and try to track it down with debug outputs.

Coryza:
Unfortunately my VCF file is huge and therefore still contains a small amount of non-decomposed MNPs - which then crashes upon providing no reference;

```
whatshap polyphase tmp.vcf Cavendish.reads.bam --ploidy 3 --ignore-read-groups --indels --mapping-quality 0 > tmp.phase.vcf
This is WhatsHap (polyploid) 1.7 running under Python 3.8.10
======== Working on chromosome 'Chr00'
---- Processing individual unknown
Number of variants skipped due to missing genotypes: 0
[...]

PR #429, opened 2023-01-20
Fixes issue 427 (polyphase not outputting indels)

schrins:
Created this for the small fix for issue #427 
We might add another change to skip MNP/substitution variants during the no-reference allele detection (with a warning) instead of crashing the program.


schrins:
Merging this pull request, as the changes are minimal and it fixes a relevant bug.

PR #432, opened 2023-01-26
Hybrid polyphase

schrins:
In this branch I experimented a bit with how the polyphase module can profit from sparse phasing information that is provided as a partially phased VCF. Such data is produced by the `polyphasegenetic` command, where only certain variant types are phased using pedigree genotype information. The research on this is not done but along the way there were also some improvements outside this experiment that should be merged back into main at some point.

1. The third stage of the polyphase algorithm was reworked. It now uses a "cleaner" formal model to reorder the results from the previous two stages. The partial pre-phasing information is considered in this stage now. In general, I expect minor accuracy improvements for polyphase, also without pre-phasing information.

2. The way phasing blocks are cut should be a bit more meaningful now. The blocks now also depend on how confident the phasing of ambiguous sites is.

3. The function in vcf.py to convert phasing blocks into a set of "superreads" was extended to polyploid phasings. Before, the readset only contained reads corresponding to one phase, because the second is always complementary. This was not possible for polyploid phasings, but for consistency I also changed the diploid case to now output superreads for both phases. I adjusted the test cases, but I am not sure if this could cause any problems.

4. During the previous step, I realized that HP phasing information was incorrectly processed for polyploids. It should now work as intended.

5. Test cases for reorder.py (the major change) and a few for other modules.

6. The docs for polyphase were updated and docs for polyphasegenetic were added.


schrins:
Bumping this pull request, because I will soon need the progress here for another branch.

I guess there is no one besides @marcelm who will raise any objections. If you find the time, it would be nice to have a look at the changes outside the polyphase module, that could potentially affect other parts of the tool. Anything that breaks inside polyphase will come back to me at some point, anyways.

marcelm (approved):
No objections!

PR #479, opened 2023-09-04
Add #include <cstdint> where needed

mp15:
The libstdc++ packaged with GCC 13.2.0 (and possibly earlier versions) requires <cstdint> to be explicitly included in places where it may have been implicitly included before.


schrins:
Thanks for pointing this out. These import sections have not been touched in a while and issues like this can stay unnoticed, if one does update certain other packages (like gcc) too often.

PR #481, opened 2023-09-10
Kmerald: a new method for allele detection based on k-mer alignment by Hufsah Ashraf.

tobiasmarschall:


tobiasmarschall:
@Hufsah-Ashraf Thanks a lot for all your work on this! Could you take a look whether you are happy with how I squashed commits and edited the commit messages? (I left the old branch for now, we should delete once merged).
@marcelm I want to merge this tomorrow. Please let me know in case you have objections (or want time to look at the code).

tobiasmarschall:
@Hufsah-Ashraf Did you write test cases for this as well? If yes it appears you haven't pushed them...

whatshap/align.pyx
```diff
@@ -191,3 +194,78 @@ def edit_distance_affine_gap(query,ref, mismatch_cost, int gap_start=1, int gap_
...
+	cdef int C = ord('C')
+	cdef int G = ord('G')
+	cdef int T = ord('T')
+	cdef c = 0
+	cdef int h = 0
+	cdef int mask = (1 << (2*k)) - 1
+	cdef int i = 0
+	cdef kmer_list= collections.deque([])
+	for i in range(len(reference)):
+		c = reference[i]
```
marcelm, line 259:
```suggestion
	for i, c in enumerate(reference):
```

whatshap/phred_scores.py
```diff
@@ -0,0 +1,89 @@
...
+    comb_count = {}
+    k = int(kmer_size)
+    total_comb = float(4**k)
+    epsilon = float(epsilon_value)
+    path = Path(inputfile)
+    glob_path = path.glob("*.txt")
+    writer = open(outputfile, "w")
+    # Iterate over each chromosome file
+    for file in glob_path:
+        counts_file = open(file, "r")
```
marcelm, line 26: Please use a context manager: `with open(file):` (to close the file automatically when no longer needed). See [the example in the csv.reader documentation](https://docs.python.org/3/library/csv.html#csv.reader).

whatshap/phred_scores.py
```diff
@@ -0,0 +1,89 @@
...
+        reader = csv.reader(counts_file, delimiter="\t")
+        for line in reader:
+            ref = line[1]
+            read = line[2]
+            count = int(line[3])
+            new_key = (ref, read)
+            if new_key in probs:
+                probs[new_key] += count
+            else:
+                probs[new_key] = count
```
marcelm, line 36: You can avoid the "check if key is in dict and then update, else set" pattern by using a [defaultdict](https://docs.python.org/3/library/collections.html#collections.defaultdict).

Initialize probs thus:
```
probs = defaultdict(int)
```
then update:
```
probs[new_key] += count
```

(Same for `seen_comb`)

whatshap/phred_scores.py
```diff
@@ -0,0 +1,89 @@
...
+    probs = {}
+    probabilities = {}
+    seen_comb = {}
+    comb_count = {}
+    k = int(kmer_size)
+    total_comb = float(4**k)
+    epsilon = float(epsilon_value)
+    path = Path(inputfile)
+    glob_path = path.glob("*.txt")
+    writer = open(outputfile, "w")
```
marcelm, line 23: Use a context manager here as well (`with open(outputfile, "w")` as writer:`)

whatshap/phred_scores.py
```diff
@@ -0,0 +1,89 @@
...
+            comb_count[every_key[0]] = probs[every_key]
+
+    for s in probs:
+        probability = probs[s] / (comb_count[s[0]] + ((total_comb - seen_comb[s[0]]) * epsilon))
+        e_probability = epsilon / (comb_count[s[0]] + ((total_comb - seen_comb[s[0]]) * epsilon))
+        phred_score = -10 * math.log10(float(probability))
+        e_phred_score = -10 * math.log10(float(e_probability))
+        if s[0] in probabilities:
+            probabilities[s[0]].append("1")
+            writer.write(str(s[0]) + "\t" + str(s[1]) + "\t" + str(phred_score) + "\n")
```
marcelm, line 55: `print()` is much nicer here:
```suggestion
            print(s[0], s[1], phred_score, sep="\t", file=writer)
```
(Same below.)

whatshap/variants.py
```diff
@@ -104,13 +110,32 @@ def __init__(
...
+            self._splitted_strings = {}
+            costs = {}
+            with open(kmeralign_costs) as costs_file:
+                reader = csv.reader(costs_file, delimiter="\t")
+                for line in reader:
+                    costs[(int(line[0]), int(line[1]))] = line[2]
+            self._kmerald_costs = costs
+
+        else:
+            self._kmerald_costs = False
```
marcelm, line 135: These "unused" attributes should be set to `None`.

whatshap/variants.py
```diff
@@ -438,25 +477,90 @@ def realign(
 
         left_cigar, right_cigar = ReadSetReader.split_cigar(cigartuples, i, consumed)
 
-        left_ref_bases, left_query_bases = ReadSetReader.cigar_prefix_length(
-            left_cigar[::-1], overhang
-        )
-        right_ref_bases, right_query_bases = ReadSetReader.cigar_prefix_length(
-            right_cigar, len(variant.reference_allele) + overhang
-        )
+        if use_kmerald:
```
marcelm, line 480: This is getting too hard to read and maintain and needs to be refactored.

whatshap/variants.py
```diff
@@ -104,13 +110,32 @@ def __init__(
...
             self._reader = SampleBamReader(paths[0], reference=reference)
         else:
             self._reader = MultiBamReader(paths, reference=reference)
 
+        if use_kmerald:
+            self._calculated_costs = {}
+            self._splitted_strings = {}
+            costs = {}
+            with open(kmeralign_costs) as costs_file:
+                reader = csv.reader(costs_file, delimiter="\t")
```
marcelm, line 129: Reading the CSV should not be done in the constructor of `ReadSetReader`.

marcelm (changes requested):
Hi, thanks for this. I have a couple of comments and it would be nice if at least some of them were addressed before this is merged. If this needs to be merged tomorrow, then this would be the minimum:
- Remove `kmerald_doc.txt`
- Change spelling of command-line options
- Fix ReST issues in the documentation
- Add a changelog entry

Also, since there are no tests, this is going to be difficult to maintain.

The `realign()` method is getting too big and needs to be refactored so that the kmerald and non-kmerald code is nicely separate. I will probably need to help with that.

tobiasmarschall:
Thanks @marcelm. The reason why I wanted to merge fast is because we want to submit a revision of a paper describing this method. But thinking about this, we should maybe decouple the two and add test cases before merging. Marcel, if you want to do more review, we would also wait before merging.

marcelm:
If there were tests, I would be ok with merging this as-is because I could just fix the remaining issues later if necessary. But as it stands, I don’t even know whether this works.

If you think the paper would benefit if kmerald is available in `main`, then you could merge now and add the tests within the next couple of days; that would be fine with me. I would also be happy to help with refactoring that I requested.

Hufsah-Ashraf:
Thanks @tobiasmarschall and @marcelm. I have tried to address them. Please let me know if you still see any issues or want any changes/additions. I agree with Tobias about holding back on the merge till the test cases are added. I will try to add the them as soon as possible.

marcelm (approved):

tobiasmarschall:
@Hufsah-Ashraf Thanks! Just one last request: Could you remove `tests/data/short-genome/learn-data/observed.txt` and have the test case generate a file in a temporary place? See for example this test for how that's done: 
https://github.com/whatshap/whatshap/blob/c7ece953dccf101cbd0f99471b122ea2188dac05/tests/test_run_haplotag.py#L10
On that note, could you rename `learn()` method in` whatshap/cli/learn.py to `run_learn()` for consistency? Also `test_learn.py` should be `test_run_learn.py`, which we are using for the integration tests of the whole run functions.

PR #494, opened 2023-10-25
fix genotype index computation for multiallelic sites

### LINKED ISSUE #473: Silent Error in Genotype::get_index()
# Issue

I found an error in how the index is produced by get_index(). I compiled the genotype.cpp code and just ran the following:

```cpp
int main() {
    vector<uint32_t> alleles = {0,2};
    Genotype test_exp = Genotype(alleles);
    cout << test_exp.get_index() << endl;
}
```

The output was 2. The correct output is 3 (since 2 is 1/1).

# Reason

The variable `gt` stores the genotype information in the reverse direction. When the object is initialized, position 0 gets the allele with the highest value and position 14 gets the lowest value. So the order is CBA.

When get_index() is called, this CBA order is used since the loop is as follows (Line 87-91 of genotype.cpp):
```cpp
for (uint32_t i = 0; i < ploidy; i++) {
	uint32_t allele = get_position(i);
	index += binomial_coefficient(k + allele - 1, allele - 1);
	k += 1;
}
```

This goes against the assumptions stated in https://genome.sph.umich.edu/wiki/Relationship_between_Ploidy,_Alleles_and_Genotypes where the order has to be ABC.

# Solution

Change line 88 to `uint32_t allele = get_position(ploidy-i-1);`

I think that should fix the issue.

# Note

This does not affect biallelic genotypes. I think that is why this went unnoticed.

jan-glx:
fixes #473

This PR is entirely due to Samarendra Pani's report in above issue. I do not understand what it does and just want to check if it solves some issues I observe that might be related.


schrins:
Sorry for overlooking this pull request. I already approved the changes, so I will merge this asap.

PR #503, opened 2023-11-22
Issue 498

schrins:
These changes fix some numeric instability in the polyphase module, where are high coverage caused some likelihood calculations to underflow to zero. The calculations were mostly moved to the log-space, with additional fallbacks for sums over likelihoods (since in log-space they can only be multiplied).


marcelm:
Forgot to mention this in my original review: Can you add a changelog entry? Fine to push directly to main if you want.

Then I think it’s time for a new release.

Issue #498, opened 2023-11-16
WhatsHap polyphase Assertion Error


NaphatCode:
WhatsHap 2.0
Run via Docker quay.io/biocontainers/whatshap:2.0--py38h2494328_0
`whatshap polyphase --ignore-read-groups -o phased.vcf --reference=chr22.fa clair3.vcf.gz sample.bam --ploidy 2`

I ran whatshap polyphase on nanopore data and got this error.

This is WhatsHap (polyploid) 2.0 running under Python 3.8.17
======== Working on chromosome 'chr22'
---- Processing individual SAMPLE
Number of variants skipped due to missing genotypes: 0
Number of remaining heterozygous variants: 229
Found 3138 reads covering 222 variants
Kept 866 reads that cover at least two variants each
Detecting connected components with weak interconnect ..
Split heterozygous variants into 14 blocks (and 0 singleton blocks).
Processing block 1 of 14 with 2 reads and 12 variants.
Processing block 2 of 14 with 4 reads and 25 variants.
Processing block 3 of 14 with 1 reads and 2 variants.
Processing block 4 of 14 with 3 reads and 23 variants.
Processing block 5 of 14 with 2 reads and 13 variants.
Processing block 6 of 14 with 1 reads and 8 variants.
Processing block 7 of 14 with 2 reads and 9 variants.
Processing block 8 of 14 with 1 reads and 32 variants.
Processing block 9 of 14 with 3 reads and 62 variants.
Processing block 10 of 14 with 842 reads and 2 variants.
No minimum in last threaded column!
Stored local cluster id was higher than size of global id vector! 0 31 4
Problem occured at position 0 in row [0, 0]
Traceback (most recent call last):
  File "/usr/local/bin/whatshap", line 10, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.8/site-packages/whatshap/__main__.py", line 64, in main
    module.main(args)
  File "/usr/local/lib/python3.8/site-packages/whatshap/cli/polyphase.py", line 549, in main
    run_polyphase(**vars(args))
  File "/usr/local/lib/python3.8/site-packages/whatshap/cli/polyphase.py", line 247, in run_polyphase
    ) = phase_single_individual(
  File "/usr/local/lib/python3.8/site-packages/whatshap/cli/polyphase.py", line 315, in phase_single_individual
    result = solve_polyphase_instance(allele_matrix, genotype_list, param, timers, prephasing)
[...]

Issue #472, opened 2023-07-30
whatshap polyphase: Assertion violated (Get != Set): XXX-XXXX -nan -nan


amvarani:
Hi there

I'm trying to phase a hexaploid plant genome showing a haploid length of ~1,6Gb
A total of six sequel II SMRT cells were mapped into a haploid version of this genome generated with hifasm + purge_dups 
The assembled haploid version of the genome is good, showing N50 > 85Mb
The BAM and VCF files were generated according to the commands below:

# minimap2 -ax map-hifi haploid.fasta HiFi.fastq.gz -t 60 |samtools view -@ 60 -Sb -| samtools sort -@ 32 -o file.sorted.bam -
# freebayes-parallel <(fasta_generate_regions.py haploid.fasta.fai 100000) 60 -f haploid.fasta -p 6 --use-best-n-alleles 6 file.sorted.bam > file.vcf

Running whatshap polyphase I have got several "Assertion violated" messages like below:

whatshap polyphase -t 30 -p 6 --ignore-read-groups -o run001.vcf.gz file.vcf file.sorted.bam

This is WhatsHap (polyploid) 2.0 running under Python 3.10.9
======== Working on chromosome 'ptg000002l_1'
---- Processing individual HiFi
Number of variants skipped due to missing genotypes: 0
Number of remaining heterozygous variants: 584422
Found 89567 reads covering 584402 variants
Kept 89501 reads that cover at least two variants each
Detecting connected components with weak interconnect ..
Split heterozygous variants into 176 blocks (and 7018 singleton blocks).
Phasing block 1 of 176 with 7112 reads and 49820 variants.
Phasing block 2 of 176 with 3996 reads and 28354 variants.
Phasing block 3 of 176 with 3612 reads and 27439 variants.
Phasing block 4 of 176 with 2972 reads and 27418 variants.
Phasing block 5 of 176 with 4031 reads and 25009 variants.
Phasing block 6 of 176 with 4479 reads and 20422 variants.
Phasing block 7 of 176 with 4696 reads and 20168 variants.
Phasing block 8 of 176 with 3512 reads and 19148 variants.
Phasing block 9 of 176 with 2937 reads and 19113 variants.
Phasing block 10 of 176 with 3019 reads and 18801 variants.
Phasing block 11 of 176 with 3622 reads and 17198 variants.
Phasing block 12 of 176 with 2891 reads and 15187 variants.
Phasing block 13 of 176 with 2726 reads and 14042 variants.
[...]

PR #511, opened 2024-01-22
Fix for remaining warnings mentioned in #472

schrins:
As written in the title, this is the second round of fixes for numerical issues when running polyphase's read scoring with very high coverage. The first round was uploaded as fix for #498 , but apparently incomplete.

1. The computation of large binomial/multinomial coefficients is now completely moved to the log space.
2. There was a certain check to detect underflows of products to zero, but it rather should have checked for overflows to inf.

In case there is a new release soon, I would like this to be in there as well.


marcelm:
I’ll merge this as soon as CI has passed and will then release 2.2 if there are no objections.

schrins:
Thanks, conflict should just be the changelog file.

Issue #516, opened 2024-02-07
Supporting supplementary alignment


tobiasmarschall:
Opening this issue to follow up on a conversation on supplementary alignments we just had with @nkkarpov @marcelm @snurk. This supersedes this historic issue, which actually has some useful thoughts and background information:
https://github.com/whatshap/whatshap/issues/23

tobiasmarschall:
Here are some first pointers: 

- The main logic to create a read set is in https://github.com/whatshap/whatshap/blob/main/whatshap/variants.py and specifically here, which is the function creating a read set: https://github.com/whatshap/whatshap/blob/cefdaececfbb1aa63176da301a0c13ff368aceb4/whatshap/variants.py#L132
- Each read is turned into a `Read` object here: https://github.com/whatshap/whatshap/blob/cefdaececfbb1aa63176da301a0c13ff368aceb4/whatshap/variants.py#L216
- I didn't check in detail, but the way paired end reads are handled might serve as a template for how to combine multiple supplementary alignments into one `Read` object, see https://github.com/whatshap/whatshap/blob/cefdaececfbb1aa63176da301a0c13ff368aceb4/whatshap/variants.py#L159

tobiasmarschall:
I don't have time to dig deeper today, but please feel free to ping us with questions.

fenderglass:
Subscribing to this! Would be super helpful for SV analysis. Are you talking about using supplementary alignments to phase variants, or haplotag supplementary alignments (independently from primary) in `whatshap haplotag`? Both are needed, and the latter is especially pressing :)

tobiasmarschall:
@fenderglass This issue was mostly about using supplementary alignments for phasing. But I'm curious: What would be a use case where you want to tag supplementary alignments independent from the primary ones? This might actually be relatively straighforward to add, but at least I never realized it was needed.

fenderglass:
@tobiasmarschall it is mostly relevant for rearrangements that connect two distant loci in one or two different chromosomes. Here are a few scenarioes, that ultimately lead to a situation when split alignments from the same read should be in different phases. Those mostly occur in cancer, but may also cause genetic disorders in germline. 

1. A distant jump within one chromosome; start and end coordinates of the jump are in to different phase blocks. Those blocks may have H1/H2 switched relative to each other. 
2. Even if there is full-chromosome phasing and everything is one phase block, it is possible that a rearrangement may connect two different germline haplotypes.
[...]

PR #517, opened 2024-02-29
support of supplementary alignments in phase

nkkarpov:
Adding support for supplementary alignments from issue https://github.com/whatshap/whatshap/issues/516 with flag --use-supplementary.

The supplementary alignment must have the same orientation as the primary and be within a distance of at most 100,000 from the primary alignment. This distance parameter can be adjusted using the --supplementary-distance option. The alignments may intersect. To form the new read from the alignments, the algorithm selects only positions where alignments fully agree with the allele of the variant and uses this allele.

### DIFF
--- a/whatshap/cli/phase.py
+++ b/whatshap/cli/phase.py
@@ -311,6 +311,8 @@ def run_whatshap(
     default_gq: int = 30,
     write_command_line_header: bool = True,
     use_ped_samples: bool = False,
+    use_supplementary: bool = False,
+    supplementary_distance_threshold: int = 100_000,
     algorithm: str = "whatshap",
 ) -> None:
     """
@@ -329,6 +331,9 @@ def run_whatshap(
     read_merging_max_error_rate -- max error rate on edge of merge graph considered
     read_merging_positive_threshold -- threshold on the ratio of the two probabilities
     read_merging_negative_threshold -- threshold on the opposite ratio of positive threshold
+    use_supplementary -- use supplementary alignments with primary
+    supplementary_distance_threshold -- distance threshold for filtering supplementary alignments
+
     max_coverage
     distrust_genotypes
     include_homozygous
@@ -347,7 +352,6 @@ def run_whatshap(
         raise CommandLineError("The hapchat algorithm cannot do pedigree phasing")
     if samples is None:
         samples = []
-
     timers = StageTimer()
     logger.info(f"This is WhatsHap {__version__} running under Python {platform.python_version()}")
     numeric_sample_ids = NumericSampleIds()
@@ -378,6 +382,8 @@ def run_whatshap(
                 ignore_read_groups,
                 mapq_threshold=mapping_quality,
                 only_snvs=only_snvs,
+                use_supplementary=use_supplementary,
+                supplementary_distance_threshold=supplementary_distance_threshold,
             )
         )
         show_phase_vcfs = phased_input_reader.has_vcfs
@@ -1140,6 +1146,12 @@ def add_arguments(parser):
     arg("--use-ped-samples", dest="use_ped_samples",
         action="store_true", default=False,
         help="Only work on samples mentioned in the provided PED file.")
+    arg("--use-supplementary", dest="use_supplementary", action="store_true", default=False,
+        help="Use also supplementary alignments (default: ignore supplementary_ alignments)")
+    arg("--supplementary-distance", metavar="DIST", dest="supplementary_distance_threshold", default=100_000,
+        help="Skip supplementary alignments further than DIST bp away from the primary alignment (default: %(default)s)")
+
+
 # fmt: on
 
 
--- a/whatshap/variants.py
+++ b/whatshap/variants.py
@@ -16,7 +16,6 @@ from .bam import SampleBamReader, MultiBamReader, BamReader
 from .align import edit_distance, edit_distance_affine_gap, kmer_align, enumerate_all_kmers
 from ._variants import _iterate_cigar, _detect_alleles
 
-
 logger = logging.getLogger(__name__)
 
 
@@ -67,6 +66,22 @@ class VariantProgress:
         return [i for i, a in enumerate(self.alleles) if 0 <= a.progress < a.length]
 
 
+@dataclass
+class AlignedRead:
+    read: Read
+    is_supplementary: bool
+    is_reverse: bool
+    reference_start: int
+    reference_end: int
+
+    def distance(self, other) -> int:
+        return max(
+            other.reference_end - self.reference_start,
+            other.reference_start - self.reference_end,
+            0,
+        )
+
+
 class ReadSetReader:
     """
     Associate VCF variants with BAM reads.
@@ -95,6 +110,8 @@ class ReadSetReader:
         kmer_size: int = 7,
         kmerald_gappenalty: float = 40,
         kmerald_window: int = 25,
+        use_supplementary: bool = False,
+        supplementary_distance_threshold: int = 100_000,
     ):
         """
         paths -- list of BAM paths
@@ -121,6 +138,8 @@ class ReadSetReader:
         self._kmerald_window = kmerald_window
         self._paths = paths
         self._reader: BamReader
+        self._use_supplementary = use_supplementary
+        self._supplementary_distance_threshold = supplementary_distance_threshold
         if len(paths) == 1:
             self._reader = SampleBamReader(paths[0], reference=reference)
         else:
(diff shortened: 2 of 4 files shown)


whatshap/variants.py
```diff
@@ -179,13 +202,60 @@ def _group_paired_reads(reads: Iterable[Read]) -> Iterator[List[Read]]:
...
+                                skip.add(variant.position)
+                        else:
+                            variants[variant.position] = variant
+                union_read = Read(primary.read.name,
+                                  primary.read.mapqs[0],
+                                  primary.read.source_id,
+                                  primary.read.sample_id,
+                                  reference_start,
+                                  primary.read.BX_tag, )
+                for k, v in variants.items():
```
marcelm, line 243: Instead of `k` and `v`, please use `position` and `variant`.

whatshap/variants.py
```diff
@@ -179,13 +202,60 @@ def _group_paired_reads(reads: Iterable[Read]) -> Iterator[List[Read]]:
...
+                for read in group:
+                    if not read.is_supplementary:
+                        if primary is not None:
+                            raise ReadSetError(
+                                f"Read name {group[0].name!r} has more than two primary alignments.")
+                        primary = read
+                if primary is None:
+                    n_skipped += 1
+                    continue
+                reference_start = primary.reference_start
```
marcelm, line 222: Can the code from here maybe be factored out into a separate function (`merge_reads` or so)? There is a bit much going on and the indentation is a bit too deep.

whatshap/variants.py
```diff
@@ -387,9 +469,9 @@ def split_cigar(cigar, i, consumed):
         else:
             left = cigar[:i]
         if consumed < middle_length:
-            right = [(middle_op, middle_length - consumed)] + cigar[i + 1 :]
```
marcelm, line 390: There are more formatting changes here and below that should not be made

marcelm (commented):
Hi, as @tobiasmarschall said, this needs some tests. If you factor out a `merge_reads` function as I suggest, that would be a good candidate for a unit test.

I noticed some formatting changes that I guess your editor introduced. Please make sure to run `tox` locally to catch these issues. You can run `black` to fix the formatting (except between the `# fmt: off` and `# fmt: on` comments where you need to undo the changes some other way).

It would also be nice with some extra type annotations. We don’t run mypy automatically, yet, but it’s still nice for readability.

nkkarpov:
Hello @marcelm, I've made the refactorings as you suggested and added unit tests for the new function. Please take a look when you have time. I found that the tests on genotyping failed on my local machine. However, these tests currently fail even if I clone from the main branch :) I didn't investigate further... I've also found that the test on haplotyping started to fail. From what I can see in the logs, some reads have two primary alignments in BAM files, which, as I understand, violates the BAM format. Is my understanding correct?

marcelm:
Thanks!

> I found that the tests on genotyping failed on my local machine. However, these tests currently fail even if I clone from the main branch :)

Commit d3b6c77abd0179d1ac5d32bc54acc1e51a4eb17f, which is the most recent commit an the `main` branch, doesn’t fail for me locally nor on CI, see <https://github.com/whatshap/whatshap/actions/runs/8324657558>, so I assume it is something in this PR. But maybe it’s related to the haplotyping tests (see below).

> I've also found that the test on haplotyping started to fail. From what I can see in the logs, some reads have two primary alignments in BAM files, which, as I understand, violates the BAM format. Is my understanding correct?

I think the problem is that your code only allows one primary alignment for paired-end reads. For paired-end reads, each end can have a primary alignment.

This part of the SAM spec is relevant (and I agree and somewhat confusing):

> For each read/contig in a SAM file, it is required that one and only one line associated with the read satisfies ‘`FLAG & 0x900 == 0`’. This line is called the *primary line* of the read.

I think the point is that it says "each read", and that the SAM spec considers R1 and R2 to be two *different* reads although they have the same QNAME (they are considered to come from the same *template*, though). So there will be a primary alignment for R1 and one for R2.

(The SAM spec allows for reads with more than two "ends", but please ignore that part.)

nkkarpov:
Thank you for the explanation! I fixed the issue and added more tests to cover this case.

whatshap/variants.py
```diff
@@ -169,7 +188,67 @@ def _make_readset_from_grouped_reads(groups: Iterable[List[Read]]) -> ReadSet:
...
+        variants = dict()
+        skip = set()
+        for read in group:
+            if read.is_reverse != primary.is_reverse:
+                continue
+            if primary.distance(read) > distance_threshold:
+                continue
+            reference_start = min(reference_start, read.reference_start)
+            for variant in read.read:
+                if variant.position in variants.keys():
```
marcelm, line 226:
```suggestion
                if variant.position in variants:
```

marcelm:
Nice, thanks. The unit tests are fine, but I wonder whether all corner cases are caught. Have you run this PR already on some real data? If it runs without crashing and maybe even gives the expected results, then I’d be more confident that the algorithm is ok.

Other than that, I think this only needs a changelog entry. I’ll wait with merging until tomorrow or so to give the other maintainers a chance to comment if they want.

nkkarpov:
I tested the latest version on the internal data, and the results appear reasonable. This option reduces the number of phase blocks. In our data, we have only one primary alignment for each query name, and in this case, it doesn't affect us at all.

whatshap/cli/phase.py
```diff
@@ -1140,6 +1146,12 @@ def add_arguments(parser):
     arg("--use-ped-samples", dest="use_ped_samples",
         action="store_true", default=False,
         help="Only work on samples mentioned in the provided PED file.")
+    arg("--use-supplementary", dest="use_supplementary", action="store_true", default=False,
+        help="Include supplementary alignments into reads")
```
marcelm, line 1150:
```suggestion
        help="Use also supplementary alignments (default: ignore supplementary_ alignments)")
```

whatshap/cli/phase.py
```diff
@@ -1140,6 +1146,12 @@ def add_arguments(parser):
...
+    arg("--supplementary-distance", dest="supplementary_distance_threshold", default=100_000,
+        help="Threshold for distance between primary and supplementary reads. (default: %(default)s)")
```
marcelm, line 1152:
```suggestion
    arg("--supplementary-distance", metavar="DIST", dest="supplementary_distance_threshold", default=100_000,
        help="Skip supplementary alignments further than DIST bp away from the primary alignment (default: %(default)s)")
```

whatshap/variants.py
```diff
@@ -180,13 +259,20 @@ def _group_paired_reads(reads: Iterable[Read]) -> Iterator[List[Read]]:
...
-            yield group
+            if len(group) > 1:
+                n_non_singleton += 1
+            read = ReadSetReader.create_read_from_group(group, distance_threshold)
+            if read is None:
+                n_skipped += 1
+            else:
+                yield [read]
+
+        logger.info(f"Number of non singleton groups: {n_non_singleton}")
```
marcelm, line 274:
```suggestion
        logger.info(f"Number of non-singleton groups: {n_non_singleton}")
```

marcelm (commented):
I did a last round of review and made some language-related changes. I’ll just add them all and see whether I can undo the changes to `doc/develop.rst` as well, then this can be merged since none of the other maintainers objected.

marcelm:
Nice, this has been on our wishlist for quite some time, thanks!

marcelm:
Note I squashed the changes into a single commit to make it a bit cleaner.

tobiasmarschall:
Hi @marcelm and @nkkarpov, I'm just returning from vacation today. Thanks for all your work on this!

Having a quick look at the code, my only comment is that I don't really love the `AlignedRead` class wrapped around `Read`. Is there a deeper reason for this, or was it just the reluctance to touch the `Read` class itself?

marcelm:
I’m a bit undecided: Making `Read` itself have these extra attributes would allow us to get rid of the wrapper, but on the other hand, the attributes are only needed until supplementary and primary alignments have been merged, so it also feels ok to temporarily wrap them. That is, `AlignedRead` could be considered private to the `variants.py` module.

tobiasmarschall:
Ok, given that there's pros/cons to both I'd vote to just merge the proposed version.

nkkarpov:
You are right; the main reason I did this is because I didn't want to touch the C++/Cython code. One possible solution off the top of my head is to make ```AlignedRead``` a subclass of the ```Read``` class.

leqi0001:
Hi,
Great feature! Is there a plan to allow supplemental alignment in polyphase function, or is it much more complicated to implement?
Thanks!

marcelm:
Hi @leqi0001, I believe this should actually be quite easy to add since both `phase` and `polyphase` use the same underlying routines to read in the alignments. Can you please open a new issue? Comments on closed issues or PRs get lost easily.

PR #538, opened 2024-05-31
Update guide.rst to clearly show --output-thread argument usage

helrick:
Previously this showed an `--output-thread=N` as the example, which when used caused whatshap to silently fail (writing a corrupted BAM)

### DIFF
--- a/doc/guide.rst
+++ b/doc/guide.rst
@@ -715,7 +715,7 @@ run ::
 
     whatshap haplotag -o haplotagged.bam --reference reference.fasta phased.vcf.gz alignments.bam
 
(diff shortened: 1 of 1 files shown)


marcelm:
Can you elaborate? The syntax `--option=value` should be totally fine to use as an alternative to `--option value`. I prefer writing it that way in the documentation to make it clear that `--option` takes an argument. (That is, that `--option` is not a flag.)

helrick:
Hi there, I've performed a test re-run on the eight BAM files with the `--output-threads=N` notation and everything has worked correctly. It must have been another unrelated issue on my end which caused the problem. Apologies!

PR #545, opened 2024-06-27
Add support of multi-allelic variants into haplotagphase

nkkarpov:
Sometimes we can have two non-reference alleles. Previously, there was no way to support them because the phase and haplotagphase commands didn't support multi-allelic variants. I need to ensure that I choose alleles only from the list provided by the GT tag. It requires small changes in the logic around allele detection.

### DIFF
--- a/tests/test_run_haplotagphase.py
+++ b/tests/test_run_haplotagphase.py
@@ -30,7 +30,7 @@ def test_nomav_haplototagphase(tmpdir):
         alignment_file="tests/data/pacbio/haplotagged.bam",
         reference="tests/data/pacbio/reference.fasta",
         output=outvcf,
-        ignore_mav=True,
+        mav=False,
     )
     tables = list(VcfReader(outvcf, phases=True, mav=True))
     for table in tables:
--- a/tests/test_run_haplotagphase.py
+++ b/tests/test_run_haplotagphase.py
@@ -50,7 +50,7 @@ def test_compute_votes():
     votes = compute_votes(
         {1: False, 2: False, 3: True},
         [a, b, c],
-        al2id={
+        allele_to_id={
             1: {0: 0, 1: 1},
             2: {0: 0, 1: 1},
             3: {0: 0, 1: 1},
--- a/tests/test_run_haplotagphase.py
+++ b/tests/test_run_haplotagphase.py
@@ -16,14 +16,14 @@ def test_haplotagphase(tmpdir):
         reference="tests/data/pacbio/reference.fasta",
         output=outvcf,
     )
-    tables = list(VcfReader(outvcf, phases=True))
+    tables = list(VcfReader(outvcf, phases=True, mav=True))
     for table in tables:
         assert len(table.phases) == 1
         n_unphased = sum(1 for phase in table.phases[0] if phase is None)
         assert n_unphased == 4
 
 
-def test_hapltotagphase(tmpdir):
+def test_nomav_hapltotagphase(tmpdir):
     outvcf = tmpdir.join("output.vcf")
     run_haplotagphase(
         variant_file="tests/data/pacbio/variants_haplotagphase.vcf",
@@ -32,9 +32,10 @@ def test_hapltotagphase(tmpdir):
         output=outvcf,
         ignore_mav=True,
     )
-    tables = list(VcfReader(outvcf, phases=True))
+    tables = list(VcfReader(outvcf, phases=True, mav=True))
     for table in tables:
         assert len(table.phases) == 1
+        print([phase for phase in table.phases[0]])
         n_unphased = sum(1 for phase in table.phases[0] if phase is None)
         assert n_unphased == 6
 
--- a/tests/test_run_haplotagphase.py
+++ b/tests/test_run_haplotagphase.py
@@ -11,7 +11,7 @@ from whatshap.vcf import VcfReader
 def test_haplotagphase(tmpdir):
     outvcf = tmpdir.join("output.vcf")
     run_haplotagphase(
-        variant_file="tests/data/pacbio/variants.vcf",
+        variant_file="tests/data/pacbio/variants_haplotagphase.vcf",
         alignment_file="tests/data/pacbio/haplotagged.bam",
         reference="tests/data/pacbio/reference.fasta",
         output=outvcf,
--- a/tests/test_run_haplotagphase.py
+++ b/tests/test_run_haplotagphase.py
@@ -47,5 +47,13 @@ def test_compute_votes():
         1: {(0, 0): 50, (0, 1): 0, (1, 1): 20, (1, 0): 0},
         2: {(0, 0): 10, (0, 1): 30},
     }
-    votes = compute_votes({1: False, 2: False, 3: True}, [a, b, c])
+    votes = compute_votes(
+        {1: False, 2: False, 3: True},
+        [a, b, c],
+        al2id={
+            1: {0: 0, 1: 1},
+            2: {0: 0, 1: 1},
+            3: {0: 0, 1: 1},
+        },
+    )
     assert votes == expected_votes
--- a/tests/test_run_haplotagphase.py
+++ b/tests/test_run_haplotagphase.py
@@ -23,7 +23,7 @@ def test_haplotagphase(tmpdir):
         assert n_unphased == 4
 
 
-def test_nomav_hapltotagphase(tmpdir):
+def test_nomav_haplototagphase(tmpdir):
     outvcf = tmpdir.join("output.vcf")
     run_haplotagphase(
         variant_file="tests/data/pacbio/variants_haplotagphase.vcf",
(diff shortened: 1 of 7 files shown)


tobiasmarschall:
Thanks, @nkkarpov. I'm CC'ing @schrins who has also dealt with multi-allelic variants in the context of polyploid phasing. Not sure whether there is synergy between them, but we should ideally work towards a consistent solution.

marcelm:
Just wanted to let you know that I’m on vacation next week, so if my review is requested, it’ll take some time.

whatshap/variants.py
```diff
@@ -302,7 +308,7 @@ def _usable_alignments(self, chromosome, sample, regions=None):
     def has_reference(self, chromosome):
         return self._reader.has_reference(chromosome)
 
-    def _alignments_to_reads(self, alignments, variants, sample, reference):
+    def _alignments_to_reads(self, alignments, variants, sample, reference, genotypes):
```
tobiasmarschall, line 311: See comment above. Would be good to have an explanation why a genotype is passed here.

tobiasmarschall:
@nkkarpov Sorry for the delay. I had a look and added a few comments.

whatshap/variants.py
```diff
@@ -167,15 +169,18 @@ def read(self, chromosome, variants, sample, reference, regions=None) -> ReadSet
             ignored and all reads in the file are used.
         reference -- reference sequence of the given chromosome (or None)
         regions -- list of start,end tuples (end can be None)
+        genotypes -- list of genotypes (or None)
```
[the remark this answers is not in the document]
nkkarpov: I renamed this variable to the `valid_alleles`. Also, I extend the explanation of what is going on.

whatshap/cli/haplotagphase.py
```diff
@@ -113,18 +117,27 @@ def run_haplotagphase(
...
+                        chromosome, variant_table.variants, sample, genotypes=genotypes
+                    )
                 phases = variant_table.phases_of(sample)
-                genotypes = variant_table.genotypes_of(sample)
+
                 homozygous = dict()
                 change = dict()
                 phased = dict()
+                al2id = defaultdict(dict)
+                id2al = defaultdict(dict)
```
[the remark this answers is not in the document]
nkkarpov: I renamed and added some comments.

whatshap/variants.py
```diff
@@ -149,7 +149,9 @@ def __init__(
     def n_paths(self) -> int:
         return len(self._paths)
 
-    def read(self, chromosome, variants, sample, reference, regions=None) -> ReadSet:
+    def read(
+        self, chromosome, variants, sample, reference, regions=None, valid_alleles=None
```
tobiasmarschall, line 153: Calling this `valid_alleles` is better. Thanks. Reading the code, there is the implicit assumption that valid_alleles has the same length as `variants`, correct? I would make this explicit by adding an `assert` and by mentioning in the docstring. Also, I would already mention that this needs to be a list of `Genotype` instance, because otherwise it will fail further down the line.

tests/test_run_haplotagphase.py
tobiasmarschall, line 1: Could you test the no-mav option as well? I.e. verify that it changes the outcome as expected?

whatshap/variants.py
```diff
@@ -574,11 +582,12 @@ def realign(
         splitted_strings,
     ):
         """
-        Realign a read to the two alleles of a single variant.
+        Realign a read to the two alleles (or to the valid_alleles if it is not None) of a single variant.
         i and consumed describe where to split the cigar into a part before the
         variant position and into a part starting at the variant position, see split_cigar().
 
         variant -- VcfVariant
+        valid_alleles -- list of valid alleles
```
tobiasmarschall, line 590: The variable name `valid_alleles` is used for different things in different places, which makes the code a bit harder to read than it needs to be. I think it is a list of `Genotype` objects above and just a single `Genotype` here. Would make it easier if it we used different names, e.g. `valid_alleles` here and `valid_allele_list` above.

tobiasmarschall:
@marcelm I added you to reviewers because this PR affects central files such as `variants.py`. I'd be ok with merging it after the comments I made have been addressed, but wanted to give you the chance to decide whether you want to review it or not before merging.

marcelm:
I was a bit busy recently. Please give me some time for a review until end of next week (Oct. 25th).

whatshap/variants.py
nkkarpov: I added type-hint and changed name to allowed_genotypes. Also added assert.

tests/test_run_haplotagphase.py
nkkarpov: Done.

whatshap/variants.py
nkkarpov: I start to use `predicted_genotypes` and `predicted_alleles`.

tests/test_run_haplotagphase.py
```diff
@@ -11,16 +11,33 @@
...
-    tables = list(VcfReader(outvcf, phases=True))
+    tables = list(VcfReader(outvcf, phases=True, mav=True))
     for table in tables:
         assert len(table.phases) == 1
         n_unphased = sum(1 for phase in table.phases[0] if phase is None)
-        assert n_unphased <= 4
+        assert n_unphased == 4
+
+
+def test_nomav_hapltotagphase(tmpdir):
```
marcelm, line 26:
```suggestion
def test_nomav_haplototagphase(tmpdir):
```

whatshap/cli/haplotagphase.py
```diff
@@ -39,6 +40,7 @@ def add_arguments(parser):
     arg("--chromosome", dest="chromosomes", metavar="CHROMOSOME", default=[], action="append",
         help="Name of chromosome to phase. If not given, all chromosomes in the input VCF are phased. "
         "Can be used multiple times.")
+    arg("--no-mav", dest="ignore_mav", default=False, action="store_true", help="Ignore multiallelic variants.")
```
marcelm, line 43: To avoid the double negative of `ignore_mav=False`, please invert the logic:
```suggestion
    arg("--no-mav", dest="mav", default=True, action="store_false", help="Ignore multiallelic variants.")
```

whatshap/cli/haplotagphase.py
```diff
@@ -56,6 +58,7 @@ def run_haplotagphase(
     gap_threshold: int = 70,
     cut_poly: int = 10,
     write_command_line_header: bool = True,
+    ignore_mav: bool = False,
```
marcelm, line 61: Same here, please change this so that it is `mav: bool = True`

whatshap/cli/haplotagphase.py
```diff
@@ -113,18 +117,27 @@ def run_haplotagphase(
             sample_to_super_reads, sample_to_components = (dict(), dict())
             for sample in vcf_reader.samples:
                 logger.info(f"Processing sample {sample}")
+                genotypes = variant_table.genotypes_of(sample)
                 with timers("read-bam"):
-                    reads, _ = phased_input_reader.read(chromosome, variant_table.variants, sample)
+                    reads, _ = phased_input_reader.read(
+                        chromosome, variant_table.variants, sample, predicted_genotypes=genotypes
```
marcelm, line 123: "predicted" is accurate in the sense of "the genotype predicted by the variant caller", but this isn’t relevant here. Did you actually mean "restricted" (in the sense of "genotypes restricted to those observed in the current sample")? Suggestions: `only_genotypes`, `restrict_genotypes`. (I think anything with `valid` or `allowed` in the number gives the wrong impression that "not allowed" genotypes will raise an error.)

whatshap/variants.py
```diff
@@ -149,7 +149,15 @@ def __init__(
...
 
-    def read(self, chromosome, variants, sample, reference, regions=None) -> ReadSet:
+    def read(
+        self,
+        chromosome,
+        variants,
+        sample,
+        reference,
+        regions=None,
+        predicted_genotypes: Optional[List[Genotype]] = None,
```
marcelm, line 159: see above: should be renamed

whatshap/cli/haplotagphase.py
```diff
@@ -113,18 +117,27 @@ def run_haplotagphase(
...
+                    )
                 phases = variant_table.phases_of(sample)
-                genotypes = variant_table.genotypes_of(sample)
+
                 homozygous = dict()
                 change = dict()
                 phased = dict()
+                # mapping of detected variants to 0/1 and reversed mappings.
+                allele_to_id = defaultdict(dict)
+                id_to_allele = defaultdict(dict)
```
marcelm, line 132: note to self: When we enable MAVs in `whatshap phase` as well, we should encapsulate this "mapping of variants to 0/1 and back" into a class.

marcelm (commented):
Thanks for the updates. I think the change to `predicted_genotypes` was not an improvement, and I’d like to ask you to choose something else (see my review comments).

Apart from a couple more very minor issues, I think this is ready to be merged I’m glad to see you implementing this because we can now use the same principle in `whatshap phase` to support MAVs. I’d like to encapsulate the functionality of "mapping alleles to 0/1 and back" a bit better in that case, but it can be left as it is in this PR.

marcelm:
Thanks!

Issue #523, opened 2024-03-19
WhatsHap polyphasegenetic throws an error about indel usage


schrins:
WhatsHap polyphase genetic is currently not usable, because the module complains about an `indels_used`. Adding an `--only-snvs` flag or the deprecated `--indels` flag does not change the behavior.

```schrinner@brandt:~$ whatshap polyphasegenetic /srv/homes/schrinner/data/geneticphasing/parents/ch05.56Mx.parents.decluttered.vcf /srv/homes/schrinner/data/geneticphasing/peds/ped50_5.txt -P /srv/homes/schrinner/data/geneticphasing/progeny/ch05.56Mx.progeny.decluttered.vcf.gz -p 4 -o test.vcf --sample Altus_470bp
Traceback (most recent call last):
  File "/srv/homes/schrinner/polyphase-pipeline/whatshap-polyphase-experiments/polyploid_human/evaluation/.snakemake/conda/d6c9a753ce0b321feabdd11be95d2e4f_/bin/whatshap", line 10, in <module>
    sys.exit(main())
  File "/srv/homes/schrinner/polyphase-pipeline/whatshap-polyphase-experiments/polyploid_human/evaluation/.snakemake/conda/d6c9a753ce0b321feabdd11be95d2e4f_/lib/python3.10/site-packages/whatshap/__main__.py", line 64, in main
    module.main(args)
  File "/srv/homes/schrinner/polyphase-pipeline/whatshap-polyphase-experiments/polyploid_human/evaluation/.snakemake/conda/d6c9a753ce0b321feabdd11be95d2e4f_/lib/python3.10/site-packages/whatshap/cli/polyphasegenetic.py", line 642, in main
    run_polyphasegenetic(**vars(args))
TypeError: run_polyphasegenetic() got an unexpected keyword argument 'indels_used'
```

PR #548, opened 2024-07-02
Polyphasegenetic compatibility indels

schrins:
Fix for #523


marcelm:
Can’t you instead just remove the `--indels` command-line argument?
https://github.com/whatshap/whatshap/blob/b2db272be7fe5ad3abbb4a96b735899e1fc4ef5b/whatshap/cli/polyphasegenetic.py#L536

schrins:
Yes, you are right. There is not even a warning for the deprecated parameter.

Issue #554, opened 2024-08-05
 Add Option to Exclude Specific Chromosomes in Phasing


nkkarpov:
I have a question about supporting a feature to exclude certain chromosomes from consideration. The main use case is to prevent WhatsHap from phasing reads in organisms with only one copy of chromosome X or Y.

I plan to implement it similarly to the `--chromosomes` option. I assume the right name for this flag would be `--exclude-chromosomes`, correct? Do you have a better name for this option, or perhaps prefer a different interface (for example, adding a `^` before the name in the list for the `--chromosomes` option)?

eesiribloom:
This would be helpful!

PR #562, opened 2024-10-04
Flag to exclude chromosomes

nkkarpov:
Added option for excluding chromosomes. See https://github.com/whatshap/whatshap/issues/554 for the motivation.

### DIFF
--- a/CHANGES.rst
+++ b/CHANGES.rst
@@ -5,6 +5,8 @@ Changes
 development version
 -------------------
 
+* :issue:`554`: Added ``--exclude-chromosome`` option (can be used multiple
+  times) to most subcommands (``phase``, ``haplotag``, ``genotype`` etc.)
 * :issue:`537`: Fixed a crash when running `haplotag` on CRAM files.
 * :pr:`545`: `haplotagphase` now supports multi-allelic variants.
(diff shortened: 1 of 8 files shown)


tobiasmarschall:
Hi @nkkarpov @nkarpov-ont (which user do you prefer to be used?), thanks for doing this, I think it's very useful. I looked through the code and have two comments:

1. The name `ChromosomeSet` for me would suggest something iterable, which it is not. Maybe `ChromosomeMask` is a better description for its functionality?
2. Could you add a small test to the corresponding `tests/test_run_<command>.py` that uses the exclude functionality? Having such things covered in tests is really important to be able to maintain this in the future.

nkkarpov:
Both accounts work perfectly.

1. Yes, you are right. What do you think about `ChromosomeFilter`? I made the changes that were needed.
2. One test was added, similar to the one for phasing specific chromosomes.

marcelm:
Closing and re-opening to trigger CI

marcelm:
Some tests fail on CI because pysam wheels aren’t available, yet, for Python 3.13. Since the tests pass locally, I’ll merge this anyway.

marcelm:
Thanks!

Issue #513, opened 2024-01-30
whatshap compare does not work with the results from beagle


Zhiiiiiaozhao:
I am using the whatshap compare to compare the results I got from beagle with the benchmark. But I got error like this:

Traceback (most recent call last):
  File "/usr/local/bin/whatshap", line 10, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/site-packages/whatshap/__main__.py", line 64, in main
    module.main(args)
  File "/usr/local/lib/python3.9/site-packages/whatshap/cli/compare.py", line 1077, in main
    run_compare(**vars(args))
  File "/usr/local/lib/python3.9/site-packages/whatshap/cli/compare.py", line 918, in run_compare
    ) = compare(
  File "/usr/local/lib/python3.9/site-packages/whatshap/cli/compare.py", line 491, in compare
    ) = compare_pair(
  File "/usr/local/lib/python3.9/site-packages/whatshap/cli/compare.py", line 557, in compare_pair
    if hamming(phasing0, phasing1) < hamming(phasing0[0], complement(phasing1[0])):
  File "/usr/local/lib/python3.9/site-packages/whatshap/cli/compare.py", line 115, in complement
    return "".join(t[c] for c in s)
  File "/usr/local/lib/python3.9/site-packages/whatshap/cli/compare.py", line 115, in <genexpr>
    return "".join(t[c] for c in s)
KeyError: '2'

`whatshap compare --tsv-pairwise beagle_pacbio_HG005.tsv --switch-error-bed beagle_pacbio_HG005.switch_error.bed --longest-block-tsv beagle_pacbio_HG005.longest_block.tsv HG005_benchmark.chr4.vcf.gz HG005_beagle_pacbio.vcf.gz`

[HG005_beagle_pacbio.vcf.gz](https://github.com/whatshap/whatshap/files/14100147/HG005_beagle_pacbio.vcf.gz)
[HG005_benchmark.chr4.vcf.gz](https://github.com/whatshap/whatshap/files/14100149/HG005_benchmark.chr4.vcf.gz)

this works well for the phased results from other tools, shapeit and eagle, of course the whatshap.

Could you please have a look at this?

tobiasmarschall:
Thanks for reporting this. I took a look and it's caused by multi-allelic sites (genotypes `1|2`, `2|0`, `2|1`). If you strip them from the VCF it works, I just tried this:
```
[...]

PR #571, opened 2024-11-01
Fix crash of compare command

nkkarpov:
Fix of https://github.com/whatshap/whatshap/issues/513


marcelm:
Thanks! Let’s go with the "easy fix" for now.

PR #582, opened 2025-02-01
Fix tests on macOS

nkkarpov:
This pull request focuses on stabilizing unit tests on macOS by adding robust likelihood comparisons, adjusting scaling factors to avoid numerical issues, and improving file comparison behavior.

Note: The run_learn function remains unstable, producing different results when run multiple times.


marcelm:
Awesome, thanks!

PR #583, opened 2025-02-06
Support of multisample vcf for haplotagphase

nkkarpov:
The pull request introduces several changes to support working with multiple samples in the `haplotagphase` command.

- Updated the `run_haplotagphase` function to accept an optional sample parameter and handle cases where no sample is specified. If a sample is not provided, Whatshap does not modify the phasing information for the sample.

### DIFF
--- a/tests/test_run_phase.py
+++ b/tests/test_run_phase.py
@@ -993,7 +993,7 @@ def test_with_read_merging(algorithm):
 
 def test_vcf_with_missing_headers(algorithm):
     # Since pysam 0.16, this type of invalid VCF is no longer accepted
-    with raises(CommandLineError):
+    with raises(ValueError):
         run_whatshap(
             phase_input_files=["tests/data/oneread.bam"],
             variant_file="tests/data/missing-headers.vcf",
--- a/tests/test_run_phase.py
+++ b/tests/test_run_phase.py
@@ -993,7 +993,7 @@ def test_with_read_merging(algorithm):
 
 def test_vcf_with_missing_headers(algorithm):
     # Since pysam 0.16, this type of invalid VCF is no longer accepted
-    with raises(ValueError):
+    with raises(CommandLineError):
         run_whatshap(
             phase_input_files=["tests/data/oneread.bam"],
             variant_file="tests/data/missing-headers.vcf",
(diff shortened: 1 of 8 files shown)


whatshap/cli/__init__.py
```diff
@@ -10,17 +10,13 @@
...
 from whatshap.variants import ReadSetReader, ReadSetError
-from whatshap.utils import IndexedFasta, FastaNotIndexedError, detect_file_format
+from whatshap.utils import CommandLineError, IndexedFasta, FastaNotIndexedError, detect_file_format
 from whatshap.core import Genotype, ReadSet
 from whatshap.vcf import VcfReader
 
 logger = logging.getLogger(__name__)
 
 
-class CommandLineError(Exception):
```
marcelm, line 20: Why did you move this to utils?

tests/test_run_phase.py
```diff
@@ -993,7 +993,7 @@ def test_with_read_merging(algorithm):
 
 def test_vcf_with_missing_headers(algorithm):
     # Since pysam 0.16, this type of invalid VCF is no longer accepted
-    with raises(CommandLineError):
+    with raises(ValueError):
```
marcelm, line 996: As discussed elsewhere: This should be reverted.

The idea is that all the `run_...` functions *only* raise CommandLineError. If any other exception type propagates out, it’s a bug.

whatshap/cli/__init__.py
nkkarpov: I tried to remove circular dependencies...

tests/test_run_phase.py
nkkarpov: Done.

whatshap/cli/__init__.py
marcelm: Can you solve the circular import some other way? (Not sure where it arises so I cannot say how.) Only code in the `cli` submodule should raise `CommandLineError`, which is why the exception is defined within `cli`.

The idea is that the code is logically split into two parts: Everything in the `cli` submodule is the "frontend" that implements the command-line interface, and the other, non-`cli` modules are the "backend". Everything in the backend should raise specific exceptions that indicate what exactly the problem is (e.g. `FastaNotIndexedError`). The frontend catches all the expected errors and converts them to `CommandLineError`s, which end up as user-visible error messages.

I’ve now added the above to the developer documentation.

Issue #587, opened 2025-03-12
The reasons of the untag reads


charliechen912ilovbash:
I am finding the reasons of the untagged reads after `whatshap haplotag`.
About 60% of my untagged reads are primary reads,
many of which cross at least one variant, and the MAPQ is 60.

Below is an untagged read information (for ML and MM tag, only partial content is shown, because it is too long).
Could you provide some possible explanation as to why this read is untagged?

I used the below parameters for haplotag:
```
whatshap haplotag --ignore-read-groups  --tag-supplementary \
        -o sample.bam \
        --output-haplotag-list   haplo.lst \
        --reference ref.fa \
        var.vcf.gz in.bam
```

```
Read group = da838b3b293f335dda4050c9ff25512f288338ea_rna004_130bps_sup@v5.1.0-3F5D97C5
Read length = 5,155bp
Flags = 0
----------------------
Mapping = Primary @ MAPQ 60
Reference span = chr18:14,747,969-14,854,218 (+) = 106,250bp
Cigar = 8S672M3925N115M158N174M...209M1388N135M248N27M135S
Clipping = Left 8 soft; Right 135 soft

s1 = 4342
s2 = 1888
ML = 0,105,45,0,0,0,0,0,0,0,0,0,3,28,0
MM = A+17596.,0,4,1,1,1,0,0,1,0,1,0,1
NM = 45
XM = [not shown]
MN = 5155
AS = 4304
sd = 116.09081
de = 0.006572396
ch = 775
rl = 199
cm = 1384
sm = 793.44464
fn = PAU98596_1d5e7a06_da838b3b_5.pod5
nn = 0
rn = 1151
tp = P
ms = 4901
ns = 170897
qs = 21.54998
ts = +
pt = 320
st = 2025-02-04T11:58:36.511+00:00
du = 42.72425
sv = pa
dx = 0
mx = 1
Hidden tags: MD, RG Location = chr18:14,848,824
Base = A @ QV 34
```

charliechen912ilovbash:
Sure!
I deleted the partial header (commands that include paths)

[read.zip](https://github.com/user-attachments/files/19205870/read.zip)

tobiasmarschall:
Thanks for sharing. Interestingly I could reproduce the problem in Whatshap 2.4
```
whatshap haplotag --ignore-read-groups  --tag-supplementary  -o tagged.bam --output-haplotag-list haplo.lst --reference hg38.fa snv.vcf.gz read.bam
[...]

PR #590, opened 2025-03-21
Fix distance function

nkkarpov:
- Fixed issue https://github.com/whatshap/whatshap/issues/587.
- Updated the expected exception in test_vcf_with_missing_headers - it seems vcf-parser now raises a different exception.
- Added ```hypothesis``` as a dev-dependency to enable comprehensive property-based testing.

### DIFF
--- a/tests/test_run_phase.py
+++ b/tests/test_run_phase.py
@@ -993,7 +993,7 @@ def test_with_read_merging(algorithm):
 
 def test_vcf_with_missing_headers(algorithm):
     # Since pysam 0.16, this type of invalid VCF is no longer accepted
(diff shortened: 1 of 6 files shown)


tests/test_run_phase.py
tobiasmarschall, line 1: Why is that change needed? Is it related to this problem with `distance`?

tobiasmarschall: I looked at this now and it seems unrelated. So let's handle it in a separate PR. I've opened PR #591 to address it. Let's merge that first.

nkkarpov: That sounds good to me.

PR #599, opened 2025-04-11
Move list of dependencies into pyproject.toml

marcelm:
(where it should be). This gets rid of the optimization of not compiling the extension modules on Read the Docs. We need to solve this some other way (if at all).


marcelm:
It turns out we don’t actually need to build WhatsHap itself to build the documentation (it seems), so no further changes are needed.

Issue #586, opened 2025-03-05
Variants phased in 2.2, but not in 2.3 and 2.4


munrosa:
Hi -
I think a bug was introduced in the 'phase' function when the codebase was updated from v 2.2 to v2.3 and the issue persists in v2.4. I'm providing my log file output for v2.2 and v2.4 below. If needed I can provide files for testing.
Thanks!

Here is my v2.2 output:
This is WhatsHap 2.2 running under Python 3.10.14
Working on 1 sample from 1 family

# Working on contig VGSC_TSI_amplicon in individual MUT011_S351
Found 2 usable heterozygous variants (0 skipped due to missing genotypes)
Found 2704 reads covering 2 variants
Kept 2664 reads that cover at least two variants each
Selected 15 most phase-informative reads covering 2 variants
Best-case phasing would result in 1 non-singleton phased block (0 singletons). 
Phasing 1 sample by solving the MEC problem ...
Largest block contains 2 variants (100.0% of accessible variants) between position 76 and 362

Here is the v2.4 output where there is no phasing:
This is WhatsHap 2.4 running under Python 3.12.9
Working on 1 sample from 1 family

# Working on contig VGSC_TSI_amplicon in individual MUT011_S351
Found 2 usable heterozygous variants (0 skipped due to missing genotypes)
Number of supplementary alignments: 0
Number of non-singleton groups: 2664
Skipped 0 groups
Found 2704 reads covering 2 variants
Kept 0 reads that cover at least two variants each
Selected 0 most phase-informative reads covering 0 variants
Best-case phasing would result in 0 non-singleton phased blocks (0 singletons). 
Phasing 1 sample by solving the MEC problem ...
No. of phased blocks: 0

nkkarpov:
If you can send a minimal example (bam, vcf, and command to reproduce it) it would be helpful for debugging.

tobiasmarschall:
Possibly the same underlying cause as Issue #587

tobiasmarschall:
@munrosa I just merged PR #592, which fixes Issue #587, so I hope your problem is fixed as well. Could you test again with the latest development version from git, see [these intructions](https://whatshap.readthedocs.io/en/latest/installation.html#installing-an-unreleased-development-version).

soumya1000:
**Request help with whatshap compare command for polyploids:**

I am trying to compare phased outputs against true phase, for simulated data of polyploids, using compare command. I generate my vcf files as I did not find vcf converter for polyploid data.
[...]

PR #600, opened 2025-04-11
Fix processing of paired-end reads

nkkarpov:
- Fixed a bug in the processing of paired-end reads
- Added tests
- Resolved https://github.com/whatshap/whatshap/issues/586


marcelm:
Awesome, good to have this fixed!

I’ll add a changelog entry after merging.

marcelm:
I have released version 2.6; seemed important enough.

PR #632, opened 2026-09-22
Ignore certain BX tags

### LINKED ISSUE #624: [whatshap haplotag] Introduce invalid barcode screening
In 10X/stLFR/TELLseq/Haplotagging linked-read chemistries, there are situations where the barcode is invalid for one reason or another. I looked through the source code to understand if `whatshap haplotag` omitted invalid barcodes when creating read clouds ([link](https://github.com/whatshap/whatshap/blob/f07423a7ba0c7609e7d4d5c73fb5e4e240057830/whatshap/cli/haplotag.py#L350)) and realized it doesnt. Including invalid barcodes that happen to be the same as read clouds introduces a false barcode association between reads, so it would be worth screening out those alignment records and outputting them unmodified, or perhaps lumping them together into a second set with which to phase but without their [incorrect] linked-read information.

Spefically, the invalid barcode formats for the technologies are:
- 10X/TELLseq: an `N` in the barcode (e.g. `ATGGGANAATCG`)
- stLFR: `0` for one of the segments (e.g. `2_0_459`)
- Haplotagging: `00` in any of the barcode segments (e.g. `A54C91B00D37`)

In theory, a single regex search can identify the presence of `N`, `[ABCD]00`, or `[0_|_0_|_0]`

marcelm:
Reads with BX tags that match a certain regular expression should not be used for assigning reads to read clouds.

For example, two reads that both have ACNNGT-1 cannot safely be assumed to come from the same haplotype.

Closes #624


pdimens:
I think that looks accurate, thanks! 

One of the key formatting concepts the BLink-seq group is trying to standardize is futureproofing barcode formatting with a complementary `VX:i` tag the specifies if the BX tag is valid (`1`) or invalid (`0`). The full spec is described [here](https://blinkseq.github.io/lastq/#specification). We are really trying to promote adoption of this so that validation is agnostic of specific barcode encoding. Would it be possible to modify this check to first look for a VX tag, and if not present, fall back to the regex search?

marcelm:
Ah, makes sense to have such a tag. If you plan to make this a [standard tag](https://samtools.github.io/hts-specs/SAMtags.pdf), then I think we can support that. See also https://github.com/samtools/hts-specs/blob/master/SAMv1.tex#L241-L250.

Since I’m working on this now, it would be fine for me if you just state that you intend it to become a standard tag before I implement the support; it doesn’t have to be accepted.

pdimens:
Thanks! We are already using it for all the data processed through harpy (www.github.com/pdimens/harpy), if that counts.

marcelm:
Hi, a question about a corner case while implementing this: How should a value for `VX` that is neither 0 nor 1 be interpreted? So if we encounter `VX:i:2`, should this be interpreted to mean that the barcode is valid, invalid, or should the same happen as if there were no `VX` tag at all?

pdimens:
Great question. There are two options, which I will leave to your discretion:

1. force an error and explain `VX:i:2` is not a recognized definition, it must be 0 or 1
2. conservatively treat it as a `0`, but print a notice to `stderr` saying a value >1 was detected and being inferred as `0` because it's not a recognized value

marcelm:
Hi, I’ve now added support for the `VX` tag (thanks for suggesting the tag in the hts-specs repo). I implemented it before you replied, so I actually chose the behavior that `VX` is ignored if its value is neither 0 nor 1. I think I’d like to leave it that way and not add any error messages. This is likely to not happen very often anyway and I’d have to add logic so that the error message is only printed the first time the problem is encountered.

